PR #31, opened 2017-10-02
Updates to collection tests

evant:
- Added array tests
- Moved map tests into their own file
- normalized the display of collections in failure messages
- fixed bug with array containsExactly

### DIFF
--- a/src/main/kotlin/assertk/assertions/array.kt
+++ b/src/main/kotlin/assertk/assertions/array.kt
@@ -31,25 +31,33 @@ fun <T> Assert<Array<T>>.hasSize(size: Int) {
 @JvmName("arrayContains")
 fun <T> Assert<Array<T>>.contains(element: Any?) {
     if (element in actual) return
-    expected("to contain ${show(element)} but was:${show(actual)}")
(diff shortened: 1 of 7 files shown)


src/main/kotlin/assertk/assertions/array.kt
```diff
@@ -31,25 +31,33 @@ fun <T> Assert<Array<T>>.hasSize(size: Int) {
 @JvmName("arrayContains")
 fun <T> Assert<Array<T>>.contains(element: Any?) {
     if (element in actual) return
-    expected("to contain ${show(element)} but was:${show(actual)}")
+    expected("to contain:${show(element)} but was:${show(actual)}")
```
nishtahir: We might want to think about moving these messages into a method or something. We're repeating this a lot here.

evant: Ideas? I wanted to make messages pretty flexible so it's easy to change them to give more helpful info.

nishtahir: As discussed, it's probably best to keep these pretty flexible. Approved 👍

Issue #17, opened 2017-07-08
Improve tests


evant:
All assertions should be tested.

- [x]  https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/any.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/array.kt
  - missing `all`
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/boolean.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/charsequence.kt
  - missing `isNullOrEmpty`
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/collection.kt
  - missing `all`
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/comparable.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/file.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/map.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/number.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/string.kt
- [x] https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/throwable.kt
  - missing `hasRooCause*`, `hasStackTraceContaining`

PR #32, opened 2017-10-04
Added MapSpec

pt2121:
added tests for map assertions. (partially address #17)


evant:
Sorry, we actually had some map tests but there we intermingled with the collection tests. I pulled them out in another pr. However, not all map may be tested. Can you check to see if any of the ones here can be carried over?

pt2121:
there're `hasSameSizeAs` and `containsAll` that are not tested. updated my PR.
Thanks!

PR #33, opened 2017-10-04
Upgrades Kotlin to v.1.1.50

LinkedList:
This is the latest version as of October 2017

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ group 'com.willowtreeapps.assertk'
 version '0.9'
 
 buildscript {
-    ext.kotlin_version = '1.1.2'
+    ext.kotlin_version = '1.1.50'
     ext.junit_platform_version = '1.0.0'
     ext.dokka_version = '0.9.15'
 


evant:
Hm, will have to think about our versioning strategy. Most people will have the version of kotlin explicitly defined in their project and if it's higher it will override this. I worry about getting into a situation where someone is surprised they are depending on a newer version than they defined.

LinkedList:
As I understand this all 1.1.X versions should be fully compatible between each other with maybe some deprecation warnings.

Please look at [https://kotlinlang.org/docs/reference/compatibility.html](https://kotlinlang.org/docs/reference/compatibility.html) if I am interpreting this wrongly..

evant:
> patch version updates (e.g. 1.1.X) are fully compatible

Yeah, the compiled jar will be fully compatible. I guess my concern is if there is an issue in the transitive dep (namely `"org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"`. The way gradle works, you'd always get the `1.1.50` version of that even if you explicitly were depending on an earlier one. Even worse, this will not not match the version of the the plugin (not sure if that is ever a problem though).

evant:
I can see a few options here and I'm not sure I'm completely happy with any of them.

1) Always update both the plugin and the stdlib to the latest 1.1.x version (what this pr proposes). Probably won't cause issues in practice, but may be surprising that a user though they were depending on 1.1.2 and got 1.1.50 for example.

2) Leave them both at an older version (what is currently happening). Won't be able to take advantage of any optimizations added to the kotlin compiler.

3) Update the plugin but not the stdlib. The jar is compiled with latest, but the stdlib is what the user has with a lower default.

4) Make the stdlib a provided dependency. Can depend on the latest but won't force anything on users by default. Since they are all compatible shouldn't be a problem, but will be a runtime error if they don't include the stdlib dependency at all.

Am I overthinking this? Maybe this isn't as much of an issue as I'm worried it is.

LinkedList:
The way I see it, because all 1.1.X versions are fully compatible, it doesn't matter if the library updates the project stdlib to 1.1.50 even if the project depends on a lower version because it will work the same.
And this only becomes a problem possibly with 1.X or with completely new versions of Kotlin.

And if the user is 100% sure, that he needs his version 1.1.2 for example, he can always exclude the dependency like so:
```
compile("com.willowtreeapps.assertk:assertk:0.9") {
    exclude group: "org.jetbrains.kotlin", module: "kotlin-stdlib"
}
```

And on another note, several projects/libraries from [https://github.com/Kotlin](https://github.com/Kotlin) are upgraded to the latest version so it probably isn't that big of an issue.

evant:
Good enough for me

PR #36, opened 2017-10-06
added assertions for InputStream: hasSameContentAs, hasNotSameContentAs

dibog:
contains hasSameContentAs and hasNotSameContentAs assertions for InputStreams

### DIFF
--- a/src/main/kotlin/assertk/assertions/inputstream.kt
+++ b/src/main/kotlin/assertk/assertions/inputstream.kt
@@ -1,0 +1,63 @@
+package assertk.assertions
+
+import assertk.Assert
+import assertk.assertions.support.expected
+import java.io.InputStream
+
+fun Assert<InputStream>.hasSameContentAs(other: InputStream) {
+    val buf1 = ByteArray(4096)
+    val buf2 = ByteArray(4096)
+
+    var size = 0L
+    while(true) w@ {
+        val len1 = actual.read(buf1)
+        val len2 = other.read(buf2)
+
+        when {
+            len1<len2 -> expected("to have the same size, but actual has size ${size+len1} which is smaller then the expected stream")
+            len1>len2 -> expected("to have the same size, but the expected stream size ${size+len2} which is larger then the acual stream")
+            len1==-1 -> return@w
+        }
+
+        for(i in 0..len2) {
+            val b1 = buf1[i]
+            val b2 = buf2[i]
+            if(b1!=b2) {
+                expected("to have the same content, but found first difference at pos ${size+i}")
+            }
+        }
+
+        size += len2
+    }
+}
+
+fun Assert<InputStream>.hasNotSameContentAs(other: InputStream) {
+    val buf1 = ByteArray(4096)
+    val buf2 = ByteArray(4096)
+
+    var size = 0L
+    var equals = true
+    while(true) {
+        val len1 = actual.read(buf1)
+        val len2 = other.read(buf2)
+
+        if(len1!=len2) {
+            equals = false
+            break
+        }
+        if(len1==-1) {
+            equals = true
+            break
+        }
+
+        if(!buf1.contentEquals(buf2)) {
+            equals = false
+            break
+        }
+    }
+
+    if(equals) {
+        expected("stream not to be equal, but they were equal")
+    }
+}
+
--- a/src/main/kotlin/assertk/assertions/inputstream.kt
+++ b/src/main/kotlin/assertk/assertions/inputstream.kt
@@ -9,14 +9,14 @@ fun Assert<InputStream>.hasSameContentAs(other: InputStream) {
     val buf2 = ByteArray(4096)
 
     var size = 0L
-    while(true) w@ {
+    while(true) {
         val len1 = actual.read(buf1)
         val len2 = other.read(buf2)
 
         when {
             len1<len2 -> expected("to have the same size, but actual has size ${size+len1} which is smaller then the expected stream")
             len1>len2 -> expected("to have the same size, but the expected stream size ${size+len2} which is larger then the acual stream")
-            len1==-1 -> return@w
+            len1==-1 -> return
         }
 
         for(i in 0..len2) {
@@ -35,8 +35,7 @@ fun Assert<InputStream>.hasNotSameContentAs(other: InputStream) {
     val buf1 = ByteArray(4096)
     val buf2 = ByteArray(4096)
 
-    var size = 0L
-    var equals = true
+    var equals: Boolean
     while(true) {
         val len1 = actual.read(buf1)
         val len2 = other.read(buf2)
--- a/src/main/kotlin/assertk/assertions/path.kt
+++ b/src/main/kotlin/assertk/assertions/path.kt
@@ -1,0 +1,56 @@
+package assertk.assertions
+
+import assertk.Assert
+import assertk.assertions.support.expected
+import assertk.assertions.support.show
+import java.nio.file.Files
+import java.nio.file.LinkOption
+import java.nio.file.Path
+
+fun Assert<Path>.isRegularFile(vararg options: LinkOption) {
+    if(!Files.isRegularFile(actual, *options)) {
+        expected("${show(actual)} to be a regular file, but it is not")
+    }
+}
+
+fun Assert<Path>.isDirectory(vararg options: LinkOption) {
+    if(!Files.isDirectory(actual, *options)) {
+        expected("${show(actual)} to be a directory, but it is not")
+    }
+}
+
+fun Assert<Path>.isExecutable() {
+    if(!Files.isExecutable(actual)) {
+        expected("${show(actual)} to be an executable, but it is not")
+    }
+}
+
+fun Assert<Path>.isHidden() {
+    if(!Files.isHidden(actual)) {
+        expected("${show(actual)} to be hidden, but it is not")
+    }
+}
+
+fun Assert<Path>.isReadable() {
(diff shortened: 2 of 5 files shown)


dibog:
added support for Path

dibog:
all tests passed in CircleCI

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
...
+import assertk.assertions.support.expected
+import java.io.InputStream
+
+fun Assert<InputStream>.hasSameContentAs(other: InputStream) {
+    val buf1 = ByteArray(4096)
+    val buf2 = ByteArray(4096)
+
+    var size = 0L
+    while(true) {
+        val len1 = actual.read(buf1)
```
evant: Don't think this test is right. `read` doesn't define how many bytes it will read each time https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[]). Check out the implementation in apache commons http://commons.apache.org/proper/commons-io/javadocs/api-release/src-html/org/apache/commons/io/IOUtils.html#line.2551

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
...
+        val len2 = other.read(buf2)
+
+        when {
+            len1<len2 -> expected("to have the same size, but actual has size ${size+len1} which is smaller then the expected stream")
```
evant: Think it would be more useful to consume the rest of the other stream and report both sizes. ex:
`expected("to have the same size, but expected was:${expectedSize} and actual was:${actualSize}")

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
...
+            len1>len2 -> expected("to have the same size, but the expected stream size ${size+len2} which is larger then the acual stream")
+            len1==-1 -> return
+        }
+
+        for(i in 0..len2) {
+            val b1 = buf1[i]
+            val b2 = buf2[i]
+            if(b1!=b2) {
+                expected("to have the same content, but found first difference at pos ${size+i}")
```
evant: Should report what the difference is as well as it's position.

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
...
+        val len1 = actual.read(buf1)
+        val len2 = other.read(buf2)
+
+        when {
+            len1<len2 -> expected("to have the same size, but actual has size ${size+len1} which is smaller then the expected stream")
+            len1>len2 -> expected("to have the same size, but the expected stream size ${size+len2} which is larger then the acual stream")
+            len1==-1 -> return
+        }
+
+        for(i in 0..len2) {
```
evant: Is this right? Isn't `..` an inclusive range?

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+fun Assert<InputStream>.hasNotSameContentAs(other: InputStream) {
+    val buf1 = ByteArray(4096)
+    val buf2 = ByteArray(4096)
+
+    var equals: Boolean
+    while(true) {
+        val len1 = actual.read(buf1)
```
evant: Same issue with `read`

src/main/kotlin/assertk/assertions/path.kt
```diff
@@ -0,0 +1,56 @@
+package assertk.assertions
+
+import assertk.Assert
+import assertk.assertions.support.expected
+import assertk.assertions.support.show
+import java.nio.file.Files
+import java.nio.file.LinkOption
+import java.nio.file.Path
+
+fun Assert<Path>.isRegularFile(vararg options: LinkOption) {
```
evant: Can you put the path assertions in a separate pr? That way one doesn't block the other.

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,62 @@
+package assertk.assertions
+
+import assertk.Assert
+import assertk.assertions.support.expected
+import java.io.InputStream
+
+fun Assert<InputStream>.hasSameContentAs(other: InputStream) {
```
evant: I don't know if it was there when you first started looking, but I added docs to all of the assertions. Can you do the same? ex: https://github.com/willowtreeapps/assertk/blob/master/src/main/kotlin/assertk/assertions/boolean.kt

src/test/kotlin/test/assertk/assertions/InputStreamSpec.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    given("an empty stream") {
+
+        it("has the same content as another empty stream") {
+            assert(emptyStream()).hasSameContentAs(emptyStream())
+        }
+
+        it("has the same content as another empty stream (using hasNotSameContentAs)") {
+            assertThatThrownBy {
+                assert( emptyStream() ).hasNotSameContentAs( emptyStream() )
```
evant: Can you fix your formatting? The current style doesn't have spaces around fun args. If you are using intellij the format code option should fix things.

src/test/kotlin/test/assertk/assertions/InputStreamSpec.kt
```diff
@@ -0,0 +1,90 @@
...
+            val copyArray = byteArray.copyOf()
+            val otherStream = ByteArrayInputStream(copyArray)
+
+            assertThatThrownBy {
+                assert( stream() ).hasNotSameContentAs( otherStream )
+            }.hasMessage("expected stream not to be equal, but they were equal")
+        }
+
+        it("has not the same content as another stream using another byte array") {
+            val otherStream = ByteArrayInputStream(randomByteArray())
```
evant: Don't think you need a random byte array. You should test 2 byte arrays with known differences, like actual shorter, expected shorter, and same size but some content different.

src/test/kotlin/test/assertk/assertions/PathSpec.kt
```diff
@@ -0,0 +1,104 @@
...
+            assert( regularFile ).isRegularFile()
+        }
+
+        it("is not a folder") {
+            assertThatThrownBy {
+                assert( regularFile ).isDirectory()
+            }.hasMessage("expected <LICENSE> to be a directory, but it is not")
+        }
+
+//        it("is not an executable") {
```
evant: Why is this commented out?

src/test/kotlin/test/assertk/assertions/PathSpec.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.given
+import org.jetbrains.spek.api.dsl.it
+import java.nio.file.Paths
+
+class PathSpec: Spek({
+
+    given("a regular file") {
+        val regularFile = Paths.get("LICENSE")
```
evant: Instead of using project files, create temp files for the test.

src/main/kotlin/assertk/assertions/path.kt
evant: Also, path isn't in android until api 26, we might want a separate java8 artifact.

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: Yep, when I forked it, I couldn't find them. I will add them.

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: Imagine a file with a lots of differences. I'm not sure that the information what is different is really useful. I could ofc show the values of the first different byte.

src/main/kotlin/assertk/assertions/path.kt
dibog: Regarding separate PRs: How do I do that now?
Regarding separate java8 artiffact: I have even less clue how to do that.

src/test/kotlin/test/assertk/assertions/InputStreamSpec.kt
dibog: k, will do that.

src/test/kotlin/test/assertk/assertions/PathSpec.kt
dibog: Because under windows files or folders are "executable" (according to Files.isEecutable()) so this test doesn't work like you would expect it

src/main/kotlin/assertk/assertions/inputstream.kt
evant: That's what I was thinking, just the first difference, isn't that what it's doing currently?

src/main/kotlin/assertk/assertions/path.kt
evant: Don't worry about the java8 artifact, still thinking about the right solution there. For a separate pr, you can branch off this one and delete the inputstream files, then delete the path files from here.

src/test/kotlin/test/assertk/assertions/PathSpec.kt
evant: Oh weird, yeah if this method is platform specific it might be hard to test.

PR #42, opened 2017-10-07
Explain doc gen steps

evant:
- Downgraded dokka to work around https://github.com/Kotlin/dokka/issues/186


evant:
Removed the gen doc step from contributing guide. On second thought, makes more sense at release.

PR #44, opened 2017-10-07
path assertions

dibog:
- added documentation
- incorporated change requests
- removed inputstream from pull request


src/test/kotlin/test/assertk/assertions/PathSpec.kt
```diff
@@ -0,0 +1,116 @@
...
+import assertk.assertions.*
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.given
+import org.jetbrains.spek.api.dsl.it
+import java.nio.file.Files
+import java.nio.file.Path
+
+class PathSpec : Spek({
+    var regularFile: Path? = null
```
evant: You can use `lateinit` to not have to `!!` everywhere.

src/test/kotlin/test/assertk/assertions/PathSpec.kt
```diff
@@ -0,0 +1,116 @@
...
+            assert(directory!!).isSameFileAs(directory!!.toAbsolutePath())
+            assert(directory!!).isSameFileAs(directory!!.toAbsolutePath())
+        }
+    }
+})
+
+
+private fun createTempDir() = Files.createTempDirectory("tempDir")
+private fun createTempFile() = Files.createTempFile("tempFile", "").also {
+    Files.newOutputStream(it).use {
```
evant: `  it.toFile().writeBytes(ByteArray(10))` is more concise imo. (I wish kotlin had similar extension functions directly on path)

src/test/kotlin/test/assertk/assertions/PathSpec.kt
dibog: AFAIK lateinit works only for members. Only starting with kotlin 1.2 you can put into methods.

PR #45, opened 2017-10-07
Inputstreams

dibog:
- added documentation
- incorporated change requests
- removed path from pull request


src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,118 @@
...
+                    // the current buffers are equals and we can try the next block
+                    size += actualRead
+
+                } else {
+                    // the first difference is in position pos
+
+                    val actualSize = size + actualRead + consume(actual)
+                    val otherSize = size + otherRead + consume(other)
+
+                    expected("to have the same content, but actual stream differs at pos $size. Actual stream: value=0x${actualBuffer[pos].toHexString()}, size=$actualSize. Other stream: value=0x${otherBuffer[pos].toHexString()}, size=$otherSize")
```
evant: You can't assume that `expected` will terminate/throw an exception. If you are using something like `assertAll` it will collect all failures, so you need to break if you don't want to continue the loop.

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,118 @@
...
+                val otherSize = consume(other) + otherRead + size
+                if(pos==null) {
+                    expected("to have the same size, but actual size (${actualSize}) differs from other size (${otherSize})")
+                }
+                else {
+                    expected("to have the same content, but actual stream differs at pos $size. Actual stream: value=0x${actualBuffer[pos].toHexString()}, size=$actualSize. Other stream: value=0x${otherBuffer[pos].toHexString()}, size=$otherSize")
+                }
+            }
+        }
+    }
```
evant: Should you close the streams?

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/** Asserts that the actual stream has a different content as the other stream.
+ *
+ * @param other whichs content is compared to the actual one
+ */
+fun Assert<InputStream>.hasNotSameContentAs(other: InputStream) {
+    try {
+        hasSameContentAs(other)
```
evant: Similar to the comment above, `expected` might now throw an exception. You could extract the compare logic into it's own helper function that throws/returns and call them from the assertions.

src/test/kotlin/test/assertk/assertions/InputStreamSpec.kt
```diff
@@ -0,0 +1,139 @@
...
+
+
+/** Create a new empty stream. */
+internal fun emptyStream(): InputStream {
+    return ByteArrayInputStream(ByteArray(0))
+}
+
+/** Every call creates a new stream, with the same content. */
+internal fun streamA(): InputStream {
+    return ByteArrayInputStream(arrayOf(0.toByte(), 10.toByte(), 20.toByte(), 30.toByte(), 40.toByte(), 50.toByte(), 60.toByte(), 70.toByte(), 80.toByte(), 90.toByte()).toByteArray())
```
evant: I think you can use `byteArrayOf()` to not have to do so many conversions.

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: Good question.

Usually I like that the code who creates the stream also closes is.
As they are not opened in the assertion method but in the test code, it should be closed there.
On the other hand it would lead to easier test code, if the test code does not need to close it.
 
What do you think?

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: I assumed that expect leaves the method with an exception. I will add an appropriate return or break.

evant: I should document this somewhere, it's important to know for anyone writing assertions.

src/main/kotlin/assertk/assertions/inputstream.kt
evant: Hm good point. I totally agree with that sentiment. However, here the assertion always consumes all of both streams so there really isn't anything to do afterwords except close them. I also worry that this is something people will forget to do and not notice it's an issue until they have a lot of tests.

```kotlin
actualStream.use {
  expectedStream.use {
    assert(actualStream).hasSameContentAs(expectedStream)
  }
}
```
seems a bit verbose. Should definitely document that these assertions do close the streams.

dibog:
The input streams will be closed. Added that fact also to the javadoc.

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -73,60 +83,72 @@ private fun compare(len: Int, actual: ByteArray, other: ByteArray): Int? {
  */
 private fun doTheStreamHaveTheSameContent(actual: InputStream, other: InputStream): String? {
 
+    val actualStream = InputStreamWithIdempotentCloseCall(actual)
```
evant: Can't you just use a finally block around the whole while?

dibog:
@evant Perhaps you can spot why the throw IllegalStateException is necessary? I'm quite sure that the
throw is not necessary as this code should be unreachable.

dibog:
It just occurred me, that I need the checksum assertions, too.
It would add them to file and path assertions, too. 
But to reuse the code I made here I would wait until this PR is merged.

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,226 @@
...
+private const val BUFFER_SIZE = 4096
+
+private fun consume(stream: InputStream): Int {
+    val buffer = ByteArray(BUFFER_SIZE)
+
+    var consumed = 0
+    while (true) {
+        val len = stream.read(buffer)
+        if (len < 0) {
+            stream.close()
```
evant: Since the stream is closed in `doTheStreamHaveTheSameContent` it doesn't have to be here right?

src/main/kotlin/assertk/assertions/inputstream.kt
```diff
@@ -0,0 +1,226 @@
...
+        expected("streams not to be equal, but they were equal")
+    }
+
+}
+
+/** Asserts that compute the MD5 hash of the stream content.
+ *
+ * @param expectedChecksum the expected checksum, every byte is separated by ':'
+ */
+fun Assert<InputStream>.hasMD5(expectedChecksum: String) = hasChecksum("MD5", expectedChecksum)
```
evant: How common is checking a hash in this format, or at all for that reason? Wondering if this is generic enough to land in the core lib. You can always have these assertions in your application.

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: Right, they are no longer required. I will remove them.

src/main/kotlin/assertk/assertions/inputstream.kt
dibog: In my case I'm testing a REST API and try to stream some data over the write. I want to verify that I got the correct file. Instead of having a copy of the file to use hasSameContentAs I could just use hasMD5, or hasSHA1, or SHA-256 for verification.
Or are you referring to exactly the hasMD5 method?

evant: Just thinking this is niche enough that it doesn't deserve to be in the core package.

dibog: Are you talking about all the checksums or just the md5?

evant: All of them.

dibog: I have a different opinion, but it is your project. I pushed the code without the checksum assertions.

evant: Feel free to create a seperate issue for that, I may reconsider if there is enough demand.

PR #47, opened 2017-10-08
Checksum assertions

dibog:
Assertions using checksums on inputstreams


dibog:
It complains about the import statement. I think that is correct one.
Also it complained about missing documentation for the test class, which all other
test classes also not have. 
Could it be that this quality check is too strict?

evant:
It's possible, I just added it and haven't gone through to properly configure it yet.

PR #48, opened 2017-10-08
Boolean spec

dibog:
added missing package statement (unfortunately this happens very easily with Kotlin)


dibog:
same problems as with the other pull request

evant:
This seems to have the checksum pr stuff as well. Can you make the change off of master so it's self contained?

PR #49, opened 2017-10-08
Add detekt and fix warnings

evant:


src/main/kotlin/assertk/table.kt
```diff
@@ -69,104 +76,183 @@ sealed class Table(internal val columnNames: Array<String>) {
 
     protected fun forAll(f: TableFun) {
         FailureContext.run(TableFailure(this)) {
-            for (i in 0..rows.size - 1) {
+            for (i in 0 until rows.size) {
```
rf43: 👍 for `until`

PR #51, opened 2017-10-09
Add `it` in docs to ensure correct package usage

rf43:
Correcting the README to show that you need to  use `it.hasMessage(String)` 

If you do not then the IDE will continuously attempt to automatically add the following import
`import org.junit.internal.matchers.ThrowableMessageMatcher.hasMessage` which may confuse some people


rf43:
Note: Using a single lambda arg to capture the error does not have the same behavior, thus you do not need to use `it.hasMessage()`

PR #63, opened 2017-11-29
Up kotlin version to 1.2

afeozzz:
https://blog.jetbrains.com/kotlin/2017/11/kotlin-1-2-released/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ group 'com.willowtreeapps.assertk'
 version '0.10-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.1.50'
+    ext.kotlin_version = '1.2.0'
     ext.junit_platform_version = '1.0.0'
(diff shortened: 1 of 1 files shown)


nishtahir:
Thanks for your contribution.

However, I think it would be best to hold off on merging this for some time because we don't want to have to force users to update their Kotlin version. From the link you shared,

> Compatibility. In Kotlin 1.2 the language and the standard library are backwards compatible (modulo bugs): if something compiled and ran in 1.0 or 1.1, it will keep working in 1.2. To aid big teams that update gradually, we provide a compiler switch that disables new features. Here is a document covering possible pitfalls.

This means assertk will continue to work with projects that upgrate to 1.2.

FaustXVI:
@nishtahir it will be a prerequisite to #75 though (and I did the update in #76 )

evant:
We plan to do this as part of the multiplatform support, so closing this.

PR #65, opened 2017-12-02
Fixed issue where you couldn't select `isEqualTo` for nullable java strings

### LINKED ISSUE #58: Support for comparing Java Strings (nullable)
Given the Java class (with platform types):
`class Parson { public String name; }`
the following code
`assert(new Person().name).isEqualTo(new Person().name)`
picks the 'isEqualTo' from _strings.kt_, which requires not null type and fails with exception.
Adding a new method for the `String?` type should resolve the problem.
`fun Assert<String?>.isEqualTo(expected: String?)`

evant:
Fixes #58

### DIFF
--- a/src/main/kotlin/assertk/assertions/string.kt
+++ b/src/main/kotlin/assertk/assertions/string.kt
@@ -19,7 +19,7 @@ fun Assert<String>.hasLineCount(lineCount: Int) {
  * @param ignoreCase true to compare ignoring case, the default if false.
  * @see [isNotEqualTo]
  */
-fun Assert<String>.isEqualTo(other: String, ignoreCase: Boolean = false) {
+fun Assert<String?>.isEqualTo(other: String?, ignoreCase: Boolean = false) {
     if (actual.equals(other, ignoreCase)) return
     fail(other, actual)
 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/assertk/assertions/string.kt
```diff
@@ -19,7 +19,7 @@ fun Assert<String>.hasLineCount(lineCount: Int) {
  * @param ignoreCase true to compare ignoring case, the default if false.
  * @see [isNotEqualTo]
  */
-fun Assert<String>.isEqualTo(other: String, ignoreCase: Boolean = false) {
+fun Assert<String?>.isEqualTo(other: String?, ignoreCase: Boolean) {
```
nishtahir: Is removing the default value a good idea? This causes the tests such as `                assert("").isNotEqualTo("test")` to invoke `Assert<T>.isNotEqualTo(expected: Any?)`.

evant: Hm, the extension method `String.isEqualTo(other, ignoreCase)` delegates to `equals()` so the behavior should be the same. But keeping the default would technically better match the signature.

Guess the question is if it's better to match the signature of the extension, or have `isEqualTo` have the same impl for all T.

PR #76, opened 2018-03-13
[WIP] Multiplateform assertk

### LINKED ISSUE #75: Make assertk platform independent
Since Kotlin can run on the JVM and Javascipt (and soon native), what about making assertk platform independent too ?

Specific platform assertion could be shipped in platform specific dependencies.

I would happily try this :)

FaustXVI:
Closes #75


FaustXVI:
Here is a first draft :)
To go futher, I'll need to move the tests to common. The bad news is that you use junit5 and spek so far and none is ready for multiplatform tests. It means I'll have to transform every tests to `kotlin.test` tests. Is it fine for you ?

FaustXVI:
I like the `@Nested` too but I am afraid I'll have to avoid it :(
I'll move the tests and it'll be done :)

FaustXVI:
I push few tests to show you how it could become so you can give me feedback before it move all tests :)

assertk-common/src/main/kotlin/assertk/assertions/support/support.kt
```diff
@@ -11,21 +11,9 @@ import assertk.fail
...
-        is Map<*, *> -> value.entries.joinToString(
-                prefix = "{",
-                postfix = "}",
-                transform = { (k, v) -> "${display(k)}=${display(v)}" })
-        is Regex -> "/$value/"
-        else -> value.toString()
-    }
-}
+expect
+@Suppress("UndocumentedPublicFunction")
```
FaustXVI: This strange syntax is needed to make `detekt` happy. But I am not happy with it :(
First, there is the split between expect and the rest of the signature that is non sense (detekt parsing problem ?).
Secondly, the method is internal and not public, I don't understand why we have the warning in the first place !

assertk-common/src/main/kotlin/assertk/table.kt
```diff
@@ -6,7 +6,7 @@ private class TableFailure(private val table: Table) : Failure {
     private val failures: MutableMap<Int, MutableList<AssertionError>> = LinkedHashMap()
 
     override fun fail(error: AssertionError) {
-        failures.getOrPut(table.index, { ArrayList() }) += error
+        failures.getOrPut(table.index, { ArrayList() }).plusAssign(error)
```
FaustXVI: This change is needed because he `+=` seems to not work in javascript

assertk-jvm/src/main/kotlin/assertk/assertions/any.kt
```diff
@@ -1,80 +1,17 @@
+@file:JvmName("AnyJVMKt")
```
FaustXVI: This is a breaking change for java users, but i don't think it'll be a problem since `assertk` isn't really used from `java`.

build.gradle
```diff
@@ -1,73 +1,57 @@
-group 'com.willowtreeapps.assertk'
-version '0.10-SNAPSHOT'
+allprojects {
+    group 'com.willowtreeapps.assertk'
+    version '0.10-SNAPSHOT'
+}
 
 buildscript {
-    ext.kotlin_version = '1.1.50'
+    ext.kotlin_version = '1.2.30'
```
FaustXVI: This update is mandatory for multi platform projects.

assertk-common/src/main/kotlin/assertk/assertions/support/support.kt
evant: Yeah noticed that on Travis. It really shouldn't need a doc since it isn't public. Maybe worth filling a detekt issue about it?

evant:
> And I'll have to get rid of AssertJ too :(
That's fine, you'll notice the junit tests weren't using it.

assertk-common/src/main/kotlin/assertk/table.kt
evant: That seems odd. Do you know if there is a reason for this or is it a bug?

assertk-jvm/src/main/kotlin/assertk/assertions/any.kt
evant: Not to worried about the break here, though just curious why this is required?

assertk-common/src/main/kotlin/assertk/assertions/support/support.kt
FaustXVI: Yes, definitely worth it !
Will do once I reproduced it with simpler code to give them :)

assertk-common/src/main/kotlin/assertk/table.kt
FaustXVI: I didn't dived strong enough into it :/ when using the `+=` I have a compilation exception. Look like a bug at compile time but not sure :/

assertk-jvm/src/main/kotlin/assertk/assertions/any.kt
FaustXVI: It's required because the file already exists in the common module. So there is a conflict between common code and java specific code.
Another solution that I like is to create another file name `JVMExtensions` or something like that.

FaustXVI:
And done :)
Most tests are moved to the common module and compile in js and java \o/
I made the "isInstanceOf" platform independant for `KClass`
Still have to make the javascript build deployable on npm/gradle for common projects.
What do you think ?

FaustXVI:
I change a bit the build so that intellij can run tests (almost) easily.
I moved the last thing that resisted moving.
That's a huge PR :/ sorry for that. I'll stop there. The next steps (like deliver it for the js platform) should be on another PR I think.
What do you think ?

assertk-jvm/src/test/kotlin/test/assertk/assertjlite.kt
```diff
@@ -0,0 +1,46 @@
+package test.assertk
+
+import assertk.catch
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.fail
+
+class Assertions {
```
evant: I'm not really a fan of this assertjlite class. Is this just a placeholder until the tests themselves are updated to get rid of the assertj dep?

assertk-jvm/src/test/kotlin/test/assertk/assertions/AnyTest.kt
```diff
@@ -2,162 +2,167 @@ package test.assertk.assertions
...
-            @Test fun `extracts jClass()`() {
-                assertEquals(BasicObject::class.java, assert(subject as TestObject).jClass().actual)
-            }
+    @Test
+    fun `props_extracts_jClass__`() {
+        assertEquals(BasicObject::class.java, assert(subject as TestObject).jClass().actual)
+    }
 
+    @Test
+    fun `isInstanceOf_kclass_same_class_passes`() {
```
evant: These don't need backticks anymore right?

assertk-jvm/src/test/kotlin/test/assertk/assertions/AnyTest.kt
```diff
@@ -2,162 +2,167 @@ package test.assertk.assertions
...
+}
 
-        @Nested inner class `props` {
+class `AnyTest_basic_object` {
 
-            @Test fun `extracts jClass()`() {
-                assertEquals(BasicObject::class.java, assert(subject as TestObject).jClass().actual)
-            }
+    @Test
+    fun `props_extracts_jClass__`() {
```
evant: Extra underscores, find/replace issue?

assertk-jvm/src/test/kotlin/test/assertk/AssertBlockSpec.kt
```diff
@@ -5,7 +5,7 @@ import assertk.assertions.isEqualTo
 import assertk.assertions.isNegative
 import assertk.assertions.isPositive
 import assertk.assertions.message
-import org.junit.jupiter.api.Test
+import org.junit.Test
```
evant: These should be kotlin test annotation right?

assertk-common/src/test/kotlin/test/assertk/assertJLite.kt
```diff
@@ -0,0 +1,46 @@
+package test.assertk
+
+import assertk.catch
+import kotlin.reflect.KClass
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class Assertions {
```
evant: I'm not really a fan of this assertjlite class. Is this just a placeholder until the tests themselves are updated to get rid of the assertj dep?

assertk-jvm/src/test/kotlin/test/assertk/assertions/AnyTest.kt
FaustXVI: Yes indeed. I'll fix it :)

assertk-jvm/src/test/kotlin/test/assertk/assertions/AnyTest.kt
FaustXVI: Kinda yes. The underscores replaced the special characters like `(` and `)` which where used here. I agree that it better to avoid trailing underscores but I wanted to discuss it with you :)
I'll change it.

assertk-jvm/src/test/kotlin/test/assertk/AssertBlockSpec.kt
FaustXVI: Can be. But when working on the java side, IntelliJ adds the junit annotation. But it's easy to exclude. Both are fine to me. Any preference ?

assertk-common/src/test/kotlin/test/assertk/assertJLite.kt
FaustXVI: Yes, it was just a way to get rid of assertj without changing too much tests.
I think we can inline it within the tests that use it. I just didn't want to change the tests while doing that huge refactoring.

evant: Yep, that makes sense

evant:
So as a heads up, we are planning on making a release soon. After that we'll focus on merging this in for a follow-up verson.

FaustXVI:
Good.
I'll squash my commits to have a better history and remove the wiper tag :)

Xavier Detant

FaustXVI:
Please let me know once the release is done.
Merging changes to that big PR is painful. Any small conflict almost ask to redo most of the work.
I'd like to avoid spending time on this as much as possible.
Thanks

evant:
We just pushed up a release yesterday, so that should be it for changes. I don't mind dealing with the merge conflicts if you don't want to redo everything.

FaustXVI:
I am OK to try to fix the merge conflict but it will definitely need a
double check.
I won't be able to look at it before 2 days. If you have time and want to,
please feel free to fix the conflict.

Xavier Detant

assertk-common/src/main/kotlin/assertk/platform.kt
```diff
@@ -0,0 +1,8 @@
+package assertk
+
+expect
```
mkobit: the `expect` modified before the KDoc seems like an odd style choice

assertk-jvm/src/main/kotlin/assertk/assertions/support/support.kt
```diff
@@ -0,0 +1,12 @@
+@file:JvmName("SupportJVMKt")
+package assertk.assertions.support
+
+actual
```
mkobit: same question here about modifier style - here it comes before the function annotations

assertk-common/src/main/kotlin/assertk/platform.kt
FaustXVI: Yes, I don't like it either :/ but that's the only way to make detekt happy :(
It worth a but for detekt I think.

FaustXVI:
Sorry I didn't warn you :( I am doing it right now and it's almost finished.
I'll update the PR soon (hopefully)

FaustXVI:
And here it is :)
Merged and splitted \o/

evant:
That's fine, most of my changes are test fixups that are still relevant. Think I will merge this then apply that on top.

evant:
@FaustXVI Ended up being too much work to try to merge your changes with mine, ended up just creating a new pr #89, can you take a look and see if you see any issues?

assertk-common/src/main/kotlin/assertk/assertions/support/ListDiffer.kt
```diff
@@ -9,11 +9,11 @@ internal object ListDiffer {
         val diff = mutableListOf<Edit>()
         backtrack(a, b) { prevX, prevY, x, y ->
             diff.add(
-                0, when {
-                    x == prevX -> Edit.Ins(prevY, b[prevY])
-                    y == prevY -> Edit.Del(prevX, a[prevX])
-                    else -> Edit.Eq(prevX, a[prevX], prevY, b[prevY])
-                }
+                    0, when {
```
nishtahir: Some formatting weirdness here?

assertk-common/src/main/kotlin/assertk/assertions/support/support.kt
```diff
@@ -11,34 +11,30 @@ import assertk.fail
...
-        is IntArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
-        is LongArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
-        is ShortArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
-        is Regex -> "/$value/"
-        else -> value.toString()
+        else -> displayPlatformSpecific(value)
     }
 }
 
+expect
```
nishtahir: should this be 

```kotlin
expect internal fun ...
```

assertk-common/src/main/kotlin/assertk/platform.kt
nishtahir: I think it's worth removing detekt at this point. It seems to be creating more friction than helping @evant

PR #78, opened 2018-03-20
Add assertion to compare only given properties

nishtahir:
Add `isEqualToComparingOnlyGivenProperties` to make comparisons on given properties of an object. 
This aims to be a typesafe version of assertj's [`isEqualToComparingOnlyGivenFields`](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#field-by-field-only)


nishtahir:
@evant Looks like Detekt isn't liking the function name. Got suggestions for a shorter one?

evant:
How about`hasEqualProperties`?

nishtahir:
I ended up going with `isEqualToWithGivenProperties`

PR #81, opened 2018-03-29
Use `@DslMarker` to delimit assertion DSL to prevent confusing code

mkobit:
Previously, the following code is possible:

```kotlin
assert("hello").all {
  isInstanceOf(String::class)
  toStringFun().isEqualTo("hello")
  prop(String::length).all {
    length().isEqualTo(5)
    isGreaterThan(4)
  }
}
```

This could lead to confusing code of "which block am I making the assertion on?"
With the addition of `@DslMarker`, that code will generate a compiler error like:

> 'fun <T : CharSequence> Assert<String>.length(): Assert<Int>' can't be called in this context by implicit receiver. Use the explicit one if necessary

If users want to continue doing this, they will need to be explicit with the receiver.

```kotlin
assert("hello").all string@ {
  isInstanceOf(String::class)
  toStringFun().isEqualTo("hello")
  prop(String::length).all {
    this@string.length().isEqualTo(5)
    isGreaterThan(4)
  }
}
```

I'm not sure if we need it anywhere else asside from the `Assert` type.

### DIFF
--- a/src/main/kotlin/assertk/assert.kt
+++ b/src/main/kotlin/assertk/assert.kt
@@ -3,9 +3,16 @@ package assertk
 import assertk.assertions.support.show
 
 /**
+ * Marks the assertion DSL.
+ */
+@DslMarker
+annotation class AssertkDsl
+
+/**
  * An assertion. Holds an actual value to assertion on and an optional name.
  * @see [assert]
  */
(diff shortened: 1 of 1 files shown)


nishtahir:
Thanks for your contribution

PR #82, opened 2018-03-29
Add support for more types in `show` and add tests to verify the beha…

mkobit:
…vior

-----

#### Notes

* I just picked output formats for `Byte`, `Long`, `Float` to sort of disambiguate them. I know I've ran into problems before trying to compare things in JVM land like `Long` and `Integer` and being confused why my assertions error messages say they are the same thing in the string output. I believe AssertJ 
* I left the `Class<*>` alone for now, but think it might be useful to change how the classes are output anyways, and I think that will need to be touched if multiplatform support is coming
* unfortunately all of the primitive array type need to be special cased

Side note, I saw `AnyTest.kt` was recently written using JUnit Jupiter instead of Spek, but the Reame says use `Spek`. After using spek my vote is definitrely for just using JUnit Jupiter engine.

### DIFF
--- a/src/main/kotlin/assertk/assertions/support/support.kt
+++ b/src/main/kotlin/assertk/assertions/support/support.kt
@@ -15,6 +15,10 @@ private fun display(value: Any?): String {
     return when (value) {
         null -> "null"
         is String -> "\"$value\""
+        is Char -> "'$value'"
+        is Byte -> "0x%02X".format(value)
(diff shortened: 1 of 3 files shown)


evant:
Sorry, the tests are kinda a mess right now. For multiplatform support they'll need to be written just using the `kotlin.test` See https://github.com/willowtreeapps/assertk/pull/76 for some work on this.

PR #84, opened 2018-03-31
 Improved diffing for containsExactly 

evant:


src/main/kotlin/assertk/assertions/list.kt
```diff
@@ -13,10 +11,30 @@ import assertk.assertions.support.show
  * ```
  */
 fun <T> Assert<List<T>>.index(index: Int, f: (Assert<T>) -> Unit) {
-    if (index in 0 until actual.size) {
-        f(assert(actual[index], "${name ?: ""}${show(index, "[]")}"))
+    if (index in 0 until actual.size)
+        {
```
nishtahir: Some strange formatting here. Might want to run `ktlint`

evant: fixed

Issue #87, opened 2018-04-03
Consider adding a "block does not throw" variant


mkobit:
AssertJ has the `assertThatCode(() -> {}).doesNotThrowAnyException();` which can be useful for 

This can sort of be done now by doing

```kotlin
assert {
  maybeThrowingMethod()
}.returnedValue {}
```

In my own project I'm just using `private fun <T> AssertBlock<T>.doesNotThrowAnyException() = returnedValue { }` 

But there might be a more idiomatic way to do it if it is desired as part of the library.

evant:
Think adding
```
assert {
  maybeThrowingMethod()
}.doesNotThrowAnyException()
```
seems reasonable

PR #88, opened 2018-04-13
Add doesNotThrowAnyException to assert block (#87)

abyu:


evant:
Project structure was moved around for multiplatform support. You can either fix it or I'll get to it at some point.

abyu:
I have applied the changes onto the new folder structure, please have a look

PR #89, opened 2018-04-14
Multiplatform Support

evant:
Merged in  #76 and made a few changes to it (mostly tests). @FaustXVI let me know if you see any issues with this.


FaustXVI:
Yes, do see some issues.
It seems like  all the changes made by https://github.com/willowtreeapps/assertk/commit/b6b5908b43253a6553a787b4ce688f44ee7a8420 are lost.
That was a quick check, the way I did it is to `git diff` our two branches. An there a lot of change ! More than simple changes on tests :)
If your changes are simple to replicate, I suggest to take my branch an apply it on it. I probably will be simpler.

evant:
Ok you caught me, I did change other things, that's why it's so hard to try to merge the 2 changesets. Can you point to what's missing by b6b5908? I still see all the changes there ex:
https://github.com/willowtreeapps/assertk/blob/6bf4c17a7a8a9df5093542b07b2f34b127bc0c6b/assertk-common/src/main/kotlin/assertk/assertions/collection.kt#L57.

I did run a git diff between our two branches and went through it, the only thing a found were some differences in `display` for what was in the generic vs platform method.

nishtahir:
Looks like there aren't any other differences here. I'll start a review so we can get this merged in.

assertk-common/build.gradle
```diff
@@ -0,0 +1,7 @@
+apply plugin: 'kotlin-platform-common'
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-common:$kotlin_version"
```
nishtahir: Should we be using `implementation` here?

FaustXVI:
I am terribly sorry to be so slow. I am very busy these days. Hopefully,
next week will be better.

Xavier Detant

evant:
No worries, thanks for the contribution!

FaustXVI:
hummmm... it look like I miss read my diff !
Indeed I don't see anything missing anymore :)
LGTM !

assertk-common/src/test/kotlin/test/assertk/CatchSpec.kt
```diff
@@ -0,0 +1,29 @@
+package test.assertk
+
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+
+class CatchSpec_an_exception_On_catch {
+    private val subject = TestException()
+    @Test
+    fun it_should_catch_and_return_a_exception() {
```
nishtahir: Not quite sure why we moved to snake case for test case names 😕

assertk-jvm/src/main/kotlin/assertk/assertions/throwable.kt
```diff
@@ -0,0 +1,78 @@
...
+ */
+fun <T : Throwable> Assert<T>.stackTrace() = prop("stack trace", { it.stackTrace.map { it.toString() } })
+
+/**
+ * Asserts the throwable's cause is an instance of the expected java class.
+ * @see [hasCauseWithClass]
+ * @see [hasRootCauseInstanceOf]
+ */
+@Deprecated(
+    message = "Use cause().isNotNull { it.isInstanceOf(jclass) } instead.",
```
nishtahir: If we're going to be bumping major with the release, should we keep deprecated APIs?

build.gradle
```diff
@@ -1,74 +1,56 @@
-group 'com.willowtreeapps.assertk'
-version '0.10'
+allprojects {
+    group 'com.willowtreeapps.assertk'
+    version '0.11-SNAPSHOT'
+}
 
 buildscript {
-    ext.kotlin_version = '1.1.50'
+    ext.kotlin_version = '1.2.30'
```
nishtahir: Probably want to bump this up while we have the chance.

build.gradle
```diff
@@ -1,74 +1,56 @@
-group 'com.willowtreeapps.assertk'
-version '0.10'
+allprojects {
+    group 'com.willowtreeapps.assertk'
+    version '0.11-SNAPSHOT'
```
nishtahir: Also probably make this like 1.0 preview or something

assertk-jvm/src/main/kotlin/assertk/assertions/throwable.kt
evant: I have plans for those, but imo should be a separate commit. This one is big enough

assertk-common/src/test/kotlin/test/assertk/CatchSpec.kt
evant: Was a comment on the other pr, but apparently fancy characters like "spaces" don't work on multiplatform

assertk-common/src/main/kotlin/assertk/assertions/support/support.kt
```diff
@@ -11,34 +11,32 @@ import assertk.fail
...
+        is DoubleArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
         is LongArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
         is ShortArray -> value.joinToString(prefix = "[", postfix = "]", transform = ::display)
-        is Regex -> "/$value/"
-        else -> value.toString()
+        else -> displayPlatformSpecific(value)
     }
 }
 
+expect
```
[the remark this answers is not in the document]
evant: I think my update of detekt will make this no longer an issue (the problematic check is disabled by default) so it should be safe to remove.

assertk-common/src/test/kotlin/test/assertk/CatchSpec.kt
FaustXVI: Indeed, spaces in function names is not supported on the javascript platform

Issue #91, opened 2018-04-16
Add isEmpty assertions to Map 


abyu:
New assertions to Map to check if it's empty. Right I have to do something like 
assert(theMap.keys).isEmpty()

instead assert(theMap).isEmpty() would read better

evant:
Agreed, this is planned.

ntw222:
Can I contribute to this issue ?

evant:
Implemented https://github.com/willowtreeapps/assertk/blob/master/assertk-common/src/main/kotlin/assertk/assertions/map.kt#L13

PR #92, opened 2018-04-20
Add emptyMap assertion #91

kubiakba:


evant:
Sorry, this was already implemented https://github.com/willowtreeapps/assertk/blob/master/assertk-common/src/main/kotlin/assertk/assertions/map.kt#L13

PR #101, opened 2018-05-12
Support opentest4j

### LINKED ISSUE #70: Consider using Ota4j exceptions (like AssertionFailedError) to better work with JUnit 5 based frameworks
See https://github.com/ota4j-team/opentest4j

One thing to consider also is _will assertk support multiplatform project_ which may also influence this decision.

evant:
Fixes #70 

Would like some feedback on this approch.


assertk-common/src/main/kotlin/assertk/failure.kt
```diff
@@ -58,23 +61,29 @@ internal class SoftFailure : Failure {
...
-                    transform = { "- ${it.message}" },
-                    separator = "\n"
-            )
+            AssertkMultipleFailuresError(errors)
         }
     }
 }
 
+// Override message with our previous behavior.
+internal class AssertkMultipleFailuresError(private val errors: List<Throwable>) :
```
evant: Is this a good idea? Will it be a problem for any tooling? Or is it worth changing the output to match the opentest4j default.

evant: https://github.com/ota4j-team/opentest4j/issues/46

evant:
Updated to use an external dep for opentest4j

PR #104, opened 2018-05-25
Add optional lambda to isInstanceOf

evant:
This allows additional assertions when the value is the asserted type.


geoffreywiseman:
I was just looking for something like this, which I see I can find in `atrium`. This isn't released yet, I guess?

evant:
It's not release yet, hopping to get something out soon!

PR #109, opened 2018-07-14
set kotlin version to 1.12.51

jschneider:


evant:
Any particular reason for this? We are trying to be fairly conservative on what version of kotlin we are depending on.

PR #112, opened 2018-08-04
Expiremental support for kotlin native

evant:
Currently it's up on my personal bintray, worth moving over to a willowtree one? Couldn't publish to central because of https://github.com/JetBrains/kotlin-native/issues/1839.

Kotllin native doesn't have any cross-compiling currently so you need access to a linux and mac to publish (and windows too if we care about that). Might be able to set something up with travis.


evant:
Going to wait for kotlin 1.3 to be released with it's new better multiplatform project model.

PR #113, opened 2018-08-04
Update a few more tests from old format

evant:


assertk-common/src/test/kotlin/test/assertk/LensAssertSpec.kt
```diff
@@ -1,90 +0,0 @@
...
-import assertk.assert
-import assertk.assertions.index
-import assertk.assertions.isEqualTo
-import assertk.assertions.key
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.assertFails
-
-
-class LensAssertSpec_a_list_of_lists_On_index() {
```
evant: We already have tests for these in `ListTest` and `MapTest`

Issue #116, opened 2018-08-09
More than one file was found with OS independent path 'META-INF/main.kotlin_module'


evant:
Getting this with `assertk-jvm:0.11`, seems to be a known issue https://youtrack.jetbrains.com/issue/KT-23574. Workaround is to do

```
android {
    packagingOptions {
        exclude '/META-INF/main.kotlin_module'
    }
}
```

as per https://youtrack.jetbrains.com/issue/KT-23695

we will probably have to update our kotlin version and release to fix it.

PR #117, opened 2018-08-11
Updated kotlin version

osamarao:
Updated the kotlin version to `1.2.60`. This should also fix https://github.com/willowtreeapps/assertk/issues/116.


evant:
Seems like version `1.2.50` is good enough for this, changed in https://github.com/willowtreeapps/assertk/pull/120

PR #123, opened 2018-08-30
Fix signum comparison for big numbers in jvm and add `collections.containsOnly` assertion

martintreurnicht:
Right now `isPositive`, `isNegative`, `isZero`, `isNotZero` does not work for `BigInteger` or `BigDecimal`. I tried several things but what i settled on is just converting numbers to it's double value, this is actually totally fine in my opinion because we don't care about precision, we just want to get the actual signum 0, 1, -1. If there was a way to just get a signum of the abstract number class we'd be done, but unfortunately it's not that simple (as far as i know).

Additionally i've added `containsOnly` assertion for `collections`

Please feel free to let me know if there is some gotchas i didn't think of.


build.gradle
```diff
@@ -4,19 +4,21 @@ allprojects {
 }
 
 buildscript {
-    ext.kotlin_version = '1.2.50'
-    ext.dokka_version = '0.9.16'
-    ext.detekt_version = '1.0.0.RC5'
+    ext.versions = [
+        kotlin: '1.2.61',
```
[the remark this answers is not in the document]
martintreurnicht: Most people that use `kotlin` upgrade almost immediately, especially because `intellij` pushes you to do it. I would argue if you want to be conservative you should rather use `java`, but i'm happy to revert this if you feel strongly about it.

build.gradle
```diff
@@ -4,19 +4,21 @@ allprojects {
 }
 
 buildscript {
-    ext.kotlin_version = '1.2.50'
-    ext.dokka_version = '0.9.16'
-    ext.detekt_version = '1.0.0.RC5'
+    ext.versions = [
```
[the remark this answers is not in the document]
martintreurnicht: No real reason, it just looked better to me and makes it easier to add other versions in the future. But i can revert no problem

assertk-jvm/src/test/kotlin/test/assertk/assertions/support/NumberSpec.kt
```diff
@@ -0,0 +1,89 @@
...
+import assertk.assertions.isPositive
+import assertk.assertions.isZero
+import org.junit.Test
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.test.assertEquals
+import kotlin.test.assertFails
+
+
+class BigDecimalSpec_a_Number_On_isZero() {
```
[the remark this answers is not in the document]
martintreurnicht: Happy to change it, i didn't like this naming but i was just following the convention

build.gradle
```diff
@@ -4,9 +4,9 @@ allprojects {
 }
 
 buildscript {
-    ext.kotlin_version = '1.2.50'
-    ext.dokka_version = '0.9.16'
-    ext.detekt_version = '1.0.0.RC5'
+    ext.kotlin_version = '1.2.61'
```
martintreurnicht: @evant I'm happy to revert `kotlin` version, but like i said. I think an argument can be made that upgrading to slowly is also a risk. I think it's ok to be conservative when upgrading to a major version like the upcoming `1.3.x` release, as there could be breaking changes. The `kotlin` team (in my opinion) is pretty conservative with releases already and many of the updates actually fix critical bugs. Let me know what you decision is, i'm ok with reverting this if you feel strongly about it

martintreurnicht:
@evant I've reverted all the versions. Please approve this asap, i think there is some really useful stuff in here and i'd like to not complicate this pull request with too many changes

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Thu Mar 08 21:49:58 EST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
martintreurnicht: Needed to upgrade to a newer version of Gradle to use `kotlin-dsl`

publish.gradle
```diff
@@ -93,7 +93,7 @@ publishing {
                 // Remove <packaging>pom</packaging> as it should be the default jar
                 def children = asNode().children()
                 def versionIndex = children.indexOf(children.find {it.name().localPart == 'packaging'})
-                children.remove(versionIndex)
+                if (versionIndex != -1) children.remove(versionIndex)
```
evant: If this isn't present anymore sounds like the fixed the bug that required this workaround. So I think this whole block can just be deleted.

assertk-common/src/test/kotlin/test/assertk/assertions/CollectionTest.kt
```diff
@@ -8,11 +8,11 @@ import kotlin.test.assertFails
 
 class CollectionTest {
     //region isEmpty
-    @Test fun isEmpty_empty_passes() {
+    @Test fun isEmptyPasses() {
```
evant: The underscores here were on purpose, it makes it easier to read when the message is long, which can be quite common in tests. Originally It'd be something like \`isEmpty passes\`(), but that doesn't work when targeting js ☹️.

martintreurnicht: Ok, i'll revert it.

evant:
@martintreurnicht can you please separate out separate features into their own prs? Be keeping each pr focused on one feature it's easier to give feedback and they don't block each other from getting merged.

martintreurnicht:
@evant I've opened a new clean PR for this change
https://github.com/willowtreeapps/assertk/pull/144

Issue #93, opened 2018-04-25
Make tests consitent


evant:
Tests have been rewritten several times now and they are all over the place. Need to consolidate them to one style. Updated tests have a class name ending in 'Test', where ones that need to be ported over end in 'Spec' (used to be using the Spek test framework). BooleanTest is a good exmaple of the desired format:

```kotlin
package test.assertk.assertions

import assertk.assert
import assertk.assertions.isFalse
import assertk.assertions.isTrue
import kotlin.test.Test
import kotlin.test.assertEquals
import kotlin.test.assertFails

class BooleanTest {
    //region isTrue
    @Test fun isTrue_true_value_passes() {
        assert(true).isTrue()
    }

    @Test fun isTrue_false_value_fails() {
        val error = assertFails {
            assert(false).isTrue()
        }
        assertEquals("expected to be true", error.message)
    }
    //endregion

    //region isFalse
    @Test fun isFalse_false_value_passes() {
        assert(false).isFalse()
    }

    @Test fun isFalse_true_value_fails() {
        val error = assertFails {
            assert(true).isFalse()
        }
        assertEquals("expected to be false", error.message)
    }
    //endregion
}
```
Classes to convert:
- [x] NumberSpec
- [x] StringSpec
- [x] FileSpec
- [x] InputStringSpec
- [x] JavaNullableStringSpec
- [x] PathSpec

PR #130, opened 2018-10-04
Updates FileSpec -> FileTest

LinkedList:
#93 

I am not sure what is the correct convention for `region` and properties. I'll happily change it if this is wrong.


evant:
What you have is good!

Issue #118, opened 2018-08-14
Add containsOnly for Collections


gabrielittner:
Collection currently has `containsAll` which can have extra elements. It would be useful to have a method that doesn't allow extra elements for Collections. Map already has one: `containsOnly`. I'd like to also use that for Sets or Lists.

List has `containsExactly` which doesn't allow extra elements, but it also enforces ordering which might not always be wanted.

martintreurnicht:
@gabrielittner I've added this in my pull request https://github.com/willowtreeapps/assertk/pull/123

osamarao:
@evant should we close this issue now?

PR #137, opened 2018-10-07
Add function containsOnly for collections

osamarao:
Asserts the collection contains only the expected elements
In response to issue #118

### DIFF
--- a/assertk-common/src/main/kotlin/assertk/assertions/collection.kt
+++ b/assertk-common/src/main/kotlin/assertk/assertions/collection.kt
@@ -81,3 +81,16 @@ fun <T : Collection<*>> Assert<T>.containsAll(vararg elements: Any?) {
     val notFound = elements.filterNot { it in actual }
     expected("to contain all:${show(elements)} some elements were not found:${show(notFound)}")
 }
+
+/**
+ * Asserts the collection contains only the expected elements
+ * @see [containsNone]
+ * @see [containsExactly]
+ */
+fun <T : Collection<*>> Assert<T>.containsOnly(vararg elements: Any?) {
+    containsAll(*elements)
+    val unexpected = actual.filterNot { elements.contains(it) }
+    if (unexpected.isEmpty())
+        return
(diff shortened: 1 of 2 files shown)


assertk-common/src/main/kotlin/assertk/assertions/collection.kt
```diff
@@ -81,3 +82,17 @@ fun <T : Collection<*>> Assert<T>.containsAll(vararg elements: Any?) {
...
 }
+
+/**
+ * Asserts the collection contains only the expected elements
+ * @see [containsNone]
+ * @see [containsExactly]
+ * @see [containsAll]
+ */
+fun <T : Collection<*>> Assert<T>.containsOnly(vararg elements: Any?) {
+    containsAll(*elements)
```
evant: Is this needed? Think it might be confusing if containsOnly shows the containsAll error message.

assertk-common/src/test/kotlin/test/assertk/assertions/CollectionTest.kt
```diff
@@ -101,4 +101,17 @@ class CollectionTest {
...
+        assert(listOf(1, 2)).containsOnly(2, 1)
+    }
+
+    @Test fun containsOnly_more_elements_fails() {
+        val error = assertFails{
+            assert(listOf(1, 2, 3)).containsOnly(2, 1)
+        }
+        assertEquals("expected to contain only:<[2, 1]> but extra elements were found:<[3]>", error.message)
+    }
+    //endregion
```
evant: Think it's worth adding a failure test case when the list doesn't have enough elements.

assertk-common/src/main/kotlin/assertk/assertions/collection.kt
osamarao: I agree. I actually noticed this during testing and I couldn't tell which message was which. Might write the implementation for `containsAll`, or is there a better idea ?

PR #138, opened 2018-10-07
Stacktrace is shown in JVM assertions that throw an error

### LINKED ISSUE #97: assert {} block should show full stacktrace
More of the exception stacktrace should be printed out for failures to make debugging easier.

For example:

`mkobit.kt`
```kotlin
import assertk.assertions.isEqualTo

fun main(args: Array<String>) {
  assertk.assert {
    someFunction()
  }.returnedValue {
    isEqualTo(2)
  }
}

fun someFunction(): Int {
  val two = 1 + 1
  someOtherFunction()
  return two
}

fun someOtherFunction() {
  throw RuntimeException("thing")
}
```
Produces
```
Exception in thread "main" java.lang.AssertionError: expected value but threw:<java.lang.RuntimeException: thing>
	at MkobitKt.main(mkobit.kt:6)
```

When just using 

```kotlin
assertk.assert(someFunction()).isEqualTo(2)
```

Gives
```
Exception in thread "main" java.lang.RuntimeException: thing
	at MkobitKt.someOtherFunction(mkobit.kt:14)
	at MkobitKt.someFunction(mkobit.kt:9)
	at MkobitKt.main(mkobit.kt:4)
```

Which is much easier to debug.

LinkedList:
Fixes #97


assertk-jvm/src/test/kotlin/test/assertk/JVMAssertBlockTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    //region thrown error
+    @Test fun returnedValue_exception_fails_with_stacktrace() {
+        val error = assertFails {
+            errorSubject.returnedValue {
+                isPositive()
+            }
+        }
+
+        assert(error.message).isNotNull {
```
evant: I know it's not as nice, but been trying to keep assertk out of the assertions that test it.

Issue #86, opened 2018-04-03
Add methods to assert on `Path` content


mkobit:
Proposal: add a couple methods to assert on either the line content or the bytes

```kotlin
Assert<Path>.lines(charset: java.nio.charset
 = Charsets.UTF_8): Assert<List<String>>

Assert<Path>.bytes(): Assert<ByteArray>
```

mkobit:
I can add these after multiplatform support is merged to avoid any more conflicts

PR #139, opened 2018-10-07
Adds Assert<Path>.lines and .bytes

LinkedList:
Implements #86


LinkedList:
I am not too sure about the duplication of the code in `.lines()` with `.containsExactly()`
Maybe the expected block generated from `diff` should be generified into a function to allow different naming of indexes and prefix?

evant:
This is a bit different than the original proposal. It suggested to add props for lines and bytes so you could do ex:
```kotlin
assert(path).lines().contains("a line")
```

LinkedList:
That completely eluded me :smile: makes sense though :+1: I'll change it..

PR #142, opened 2018-10-08
Change all, and assertAll to take a list of assertion lambdas

evant:
Removed the concept of 'soft assertions' that don't throw, instead all
assertions throw, and `all` and `assertAll` loop through lambda with a
try/catch. This removes the foot-gun of an assertion maybe running the
rest of the body when it fails, an allows props like `index` and `key`
to return a result and fail with an exception instead of having to take
a lambda.

This has been done in a way to not cause existing code to fail to
compile. Instead, you may not see all exceptions thrown. Deprecated
functions have been kept to guide migration.

### DIFF
--- a/assertk-common/src/main/kotlin/assertk/assert.kt
+++ b/assertk-common/src/main/kotlin/assertk/assert.kt
@@ -1,6 +1,7 @@
 package assertk
 
 import assertk.assertions.support.show
+import com.willowtreeapps.opentest4k.MultipleFailuresError
 
 /**
  * Marks the assertion DSL.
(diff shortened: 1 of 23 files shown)


evant:
Would love some feedback on this, there's some pros/cons to this approach.

# pros:
* Consistent execution model for assertions, ie `fail` and `expected` always throws.
* Failable assertion functions can return a more-specific value ex: `isNotNull()` can return `Assert<T>` for chaining. This means lambdas are not required for these.

Instead of:
```kotlin
assert(foo).isNotNull { contains("bar") }
```

We can have:
```kotlin
assert(foo).isNotNull().contains("bar")
```

* Behavior is consistent between an assertion failure and other exceptions being thrown. They will both stop execution in the lambda and be collected by `assertAll` in the same way.

# cons:
`assertAll`/`all` syntax is uglier

before:
```kotlin
assertAll {
   assert(foo).isEqualTo("one")
   assert(bar).isEqualTo("two")
}
```

after:
```kotlin
assertAll({
    assert(foo).isEqualTo("one")
}, {
    assert(bar).isEqualTo("two")
})
```

This likely means these features will be less-used.

PR #144, opened 2018-10-09
Fix signum comparison for big numbers

martintreurnicht:
Right now `isPositive`, `isNegative`, `isZero`, `isNotZero` does not work for `BigInteger` or `BigDecimal`. I tried several things but what i settled on is just converting numbers to it's double value, this is actually totally fine in my opinion because we don't care about precision, we just want to get the actual signum 0, 1, -1. If there was a way to just get a signum of the abstract number class we'd be done, but unfortunately it's not that simple (as far as i know).


assertk-jvm/src/test/kotlin/test/assertk/assertions/support/NumberSpec.kt
```diff
@@ -0,0 +1,89 @@
...
+import assertk.assertions.isPositive
+import assertk.assertions.isZero
+import org.junit.Test
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.test.assertEquals
+import kotlin.test.assertFails
+
+
+class BigDecimalSpecNumberOnIsZero {
```
evant: Can you update the tests? should be able to just add them in `NumberTest`

martintreurnicht: They can't be moved to `NumberTest`, that test is environment agnostic and doesn't support `BigDecimal` or `BigInteger`

evant: In that case, make `JavaNumberTest`.

martintreurnicht: Done

Issue #98, opened 2018-05-02
Feature Request: Add ability to do quantitative predicates for groups of objects


mkobit:
I don't know the right way to describe the feature, but this is something I use all the time from AssertJ.

Some from [`ObjectEnumerableAssert`](https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/ObjectEnumerableAssert.html) to demonstrate what I'm looking for:

* `areAtLeast(int n, Condition<? super ELEMENT> condition)`
* `areAtMost(int n, Condition<? super ELEMENT> condition)`
* `areExactly(int n, Condition<? super ELEMENT> condition)`
* `areNot(Condition<? super ELEMENT> condition)`
* `are(Condition<? super ELEMENT> condition)`

evant:
Merged in `atLeast` with https://github.com/willowtreeapps/assertk/commit/87f01511b848529beec055feea4d1254c12ef3d7 think the pattern should work well for the similar methods.

evant:
So `atLeast`, `atMost`, and `exactly` are implemented now. `each` has the same semantics as `are` and you can invert the assertions in `each` to get `areNot` ex:

```kotlin
assertThat(list).each {
   it.isNotEqualTo("foo")
}
```

this seem to cover these use-cases reasonably? I'm thinking we can close this issue now.

PR #147, opened 2018-10-25
Add quantitative predicate based assertion for 'areAtLeast'

sowmyav24:
[Issue Reference](https://github.com/willowtreeapps/assertk/issues/98)

In this PR, I have used the structure used in `iterable each assertion` . Among the features listed in the issue, `areAtLeast` is implemented.  The documentation and tests are added for the same.

If this structure is acceptable, the others can be added similarly.


sowmyav24:
In the function definition, we have changed the method signature to take predicate as non-optional parameter. 
If this works, we can follow similar implementation for other features.

evant:
Thanks for the contribution! I'd like to propose a different interface though:
```kotlin
fun <E, T : Iterable<E>> Assert<T>.atLeast(times: Int, f: (Assert<E>) -> Unit) {
   ...
}
```

which would be used like
```kotlin
assert(listOf(true, false, true)).atLeast(2) { item -> item.isTrue() }
```

# Pros
* Can reuse existing assertions and their failure messages
* Similar to `each`

# Cons
* A bit more complicated to implement, we need a way for a failure to sometimes not fail
* If there isn't an existing assertion you need to create one, instead of simply doing something inline

(changed the name to `atLeast` instead of `areAtLeast` because I think it reads a bit better when doing it this way, but could go either way on this)

Thoughts?

sowmyav24:
Thanks for the inputs. I have a few considerations though:

- In cases of in-built Comparable present in assertk returning an Unit would work,  
 `Example: item.isLesserThan(any)` , `item.isGreaterThan(any)`, 

- But in cases of user defined functions, we would be forcing the predicate defined to adhere to the above signature. 
`Example : divisibleBy5(), hasItems(), contains()`

We need to accommodate comparables in the existing implementation or come up with something that supports both. 

- Regarding your point `we need a way for a failure to sometimes not fail`, Could you give an example for the same ?

- `atLeast` sound better .

evant:
> But in cases of user defined functions, we would be forcing the predicate defined to adhere to the above signature.

Yep that is a downside. Think that could be mitigated with a predicate assertion ex:

```kotlin
val divisibleByFive = { it % 5 == 0 }
assert(n).is(divisibleByFive)
```
(feedback welcome on naming this)

> Regarding your point we need a way for a failure to sometimes not fail, Could you give an example for the same?

Just talking about the implementation here. Right now if an assertion fails it always fails a test. We'd only want `n` not failing to fail the test in this case. Should be able to repurpose soft asserts for this so shouldn't be too difficult.

sowmyav24:
- As discussed above, I have added a simple predicate assertion which can be re-used when we adopt the new signature.

- Changed the signature to : 
`fun <E, T : Iterable<E>> Assert<T>.atLeast(times: Int, f: (Assert<E>) -> Unit) {
   ...
}`

-  >  Right now if an assertion fails it always fails a test. We'd only want n not failing to fail the test in this case. Should be able to repurpose soft asserts for this so shouldn't be too difficult.

For the above, have not re-used the soft asserts. Right now have incorporated in `atLeast` method itself. Need to have a look on how we can use soft asserts to consolidate the error messages into a single customised error message and re-use it

evant:
Created a pr with the soft assert thing I was mentioning, let me know what you think https://github.com/willowtreeapps/assertk/pull/156

sowmyav24:
The soft assert helps 👍 

One small suggestion: When the test fails,

>   val error = assertFails {
            assert(listOf(1, 2, 3)).atLeast(2) { it -> it.isGreaterThan(2) }
        }
        assertEquals(
            """expected to pass at least 2 times (2 failures)
            |${"\t"}expected [[0]] to be greater than:<2> but was:<1> ([1, 2, 3])
            |${"\t"}expected [[1]] to be greater than:<2> but was:<2> ([1, 2, 3])
        """.trimMargin(), error.message
        )

We display error message for each index position. This  helps in case of `assertAll` as we run assertion for each value. But this being a single test, was wondering if multiple failure messages are necessary. 

But anyhow, this also seems to work by giving detailed information  👍

evant:
You might have different error messages for different items depending on how they fail, especially if you have some kind of chaning like `it.isNotNull().itIsGreaterThan(5)`. So I do think it's worth showing the error message for each one. Do need to think of ways to clean up that output though, ex: each line doesn't need to show the whole list.

PR #151, opened 2018-10-30
Add isClose function for Number Type

dineshba:
Added `isCloseTo` compareable function.

Usage Example:
```kotlin
assert(10.0f).isCloseTo(10.0f, 0.0)
assert(10L).isCloseTo(8, 3)
assert(10.0).isCloseTo(15.0f, delta = 3)
```
It will work for any `Number` type (Int, Double, Float, Short, Long)


assertk-common/src/main/kotlin/assertk/assertions/comparable.kt
```diff
@@ -61,3 +61,14 @@ fun <A, B : Comparable<A>> Assert<B>.isStrictlyBetween(start: A, end: A) {
     if (actual > start && actual < end) return
     expected("to be strictly between:${show(start)} and ${show(end)} but was:${show(actual)}")
 }
+
+/**
+ * Asserts the value if it is close to the expected value with given delta.
+ */
+fun <T: Number> Assert<T>.isCloseTo(value: T, delta: T) {
```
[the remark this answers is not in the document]
dineshba: I agree that it very common for `float and doubles`. I thought as a `library` we can give for all `Number`(s), users can use it accordingly.

Thoughts? I am perfectly good to have this only for `float` and `double`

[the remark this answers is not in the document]
evant: So actually, don't think this will be correct for longs as converting them to a double is a narrowing conversion. https://stackoverflow.com/a/51915530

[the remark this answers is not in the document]
dineshba: Got it. Thanks. I will try to summerize
```kotlin
val longVariable = 4999999999999999999L
val doubleVariable = 5000000000000000000.0
assert(longVariable).isCloseTo(doubleVariable, delta = 0)
```
the above test should fail. But it is not because of `narrowing conversion`.

Hence we should not convert into `double`. In that case we have to write `isCloseTo` for all `primitive types`. May be we can stick to `Double` and `Float` now.

Is that okay?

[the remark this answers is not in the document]
dineshba: Added `isCloseTo` only for `Double` and `Float`.  And also change the error message like below

```kotlin
assertk.assert(10.1f).isCloseTo(15.0f, 3.5f)
// expected <10.1f> to be close to <15.0f> with delta of <3.5f>, but was not
```

Is that okay?

assertk-js/src/test/kotlin/test/assertk/assertions/ComparableJsTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertFails
+
+class ComparableJvmTest {
+    @Test
+    fun isCloseToFloat_with_delta_out_of_range_fails() {
+        val error = assertFails {
+            assert(10.1f).isCloseTo(15.0f, 3f)
+        }
+        assertEquals("expected <10.1> to be close to <15> with delta of <3>, but was not", error.message)
```
evant: If you want to put this in the common module you could do
```kotlin
assertEquals("expected ${show(10.1f)} to be close to ${show(15.0f)} with delta of ${show(3f)}, but was not", error.message)
```

assertk-common/src/main/kotlin/assertk/assertions/comparable.kt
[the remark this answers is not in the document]
evant: Yeah, that looks good!

assertk-js/src/test/kotlin/test/assertk/assertions/ComparableJsTest.kt
dineshba: We cannot move this into common. As the error messages are platform specific.
For example
```kotlin
assert(10.1f).isCloseTo(15.0f, 3.5f)
```
jvm will show error message as (with `f` for float)
```kotlin
expected <10.1f> to be close to <15.0f> with delta of <3.5f>, but was not
```
js will show error message as (without `f` for float and `15.0` will be displayed as `15`)
```kotlin
expected <10.1> to be close to <15> with delta of <3.5>, but was not
```
So I put the failure tests in jvm and js.

evant: by putting `show()` in the expected message, it should convert it to the platform-specific display.

dineshba: Yeah, I agree. I feel it is better to use `show` while showing an error message.
As we are using `show`, we have to write test specific to platform.

Am i right? Am i missing something?

evant: Sorry, think we are talking past each other. What I'm trying to say is you can put the test in the common module by using show. Ex:
```kotlin
@Test
fun isCloseToFloat_with_delta_out_of_range_fails() {
    val error = assertFails {
        assert(10.1f).isCloseTo(15.0f, 3f)
    }
    assertEquals("expected ${show(10.1f)} to be close to ${show(15.0f)} with delta of ${show(3f)}, but was not", error.message)
}
```

dineshba: Sorry. got it now. Moved tests into common module. Thanks

Issue #150, opened 2018-10-30
Make containsAll on map have a better message.


osamarao:
Currently the message from `Assert<Map<K,V>>.containsAll()` only tells which pairs were not found. It would be better if the message could enumerate what pair was expected and what it actually was.

madhumithap:
Raised a [pull request](https://github.com/willowtreeapps/assertk/pull/152)

madhumithap:
We can close this issue as the changes have been merged to master.

PR #152, opened 2018-10-30
Change assertion message for containsAll on Map

madhumithap:
This pull request changes the assertion message of `Assert<Map<K,V>>.containsAll()` to display actual map and expected map of elements. Currently it displays expected map and pair of elements that were not found.
This is in response to [https://github.com/willowtreeapps/assertk/issues/150](url)

### DIFF
--- a/assertk-common/src/main/kotlin/assertk/assertions/map.kt
+++ b/assertk-common/src/main/kotlin/assertk/assertions/map.kt
@@ -84,7 +84,8 @@ fun <K, V> Assert<Map<K, V>>.contains(element: Pair<K, V>) {
 @PlatformName("mapContainsAll")
 fun <K, V> Assert<Map<K, V>>.containsAll(vararg elements: Pair<K, V>) {
     if (elements.all { (k, v) -> actual[k] == v }) return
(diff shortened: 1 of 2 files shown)


assertk-common/src/main/kotlin/assertk/assertions/map.kt
```diff
@@ -84,8 +84,7 @@ fun <K, V> Assert<Map<K, V>>.contains(element: Pair<K, V>) {
 @PlatformName("mapContainsAll")
 fun <K, V> Assert<Map<K, V>>.containsAll(vararg elements: Pair<K, V>) {
     if (elements.all { (k, v) -> actual[k] == v }) return
-    val notFound = elements.filterNot { (k, v) -> actual[k] == v }
-    expected("to contain all:${show(elements.toMap())} some elements were not found:${show(notFound.toMap())}")
+    expected("to contain all:${show(elements.toMap())} but was:${show(actual)}")
```
evant: Hm, this just prints out both maps, think we can do better and show what's different.

madhumithap: @evant Changed the assertion message to print expected map, actual map and what's different.

PR #154, opened 2018-11-03
Wrap custom asserts in a `given` function.

evant:
Wrap custom asserts in a `given` function.

This is an alternative to #142, allowing assertions to be chained
instead of requiring lambdas. Soft assertions are still supported the
same way they were before. However, `fail()` and `expect() now always
throw exceptions which removes the foot-gun of them sometimes not.

Custom assertions that work on the actual value need to be wrapped in a
`given` function:
```kotlin
fun <T> Assert<T>.isEqualTo(expected: Any?) {
   fail(expected, actual)
}
```
to
```kotlin
fun <T> Assert<T>.isEqualTo(expected: Any?) = given { actual ->
   fail(expected, actual)
}
```

One nice thing about this `given` function is it can also be used in
line for one-off assertions.
```kotlin
assert(foo).given { actual ->
    if (actual >= 10)
    	expected("to be less than 10 but was:${show(actual)}")
}
```

### DIFF
--- a/assertk-common/src/main/kotlin/assertk/assert.kt
+++ b/assertk-common/src/main/kotlin/assertk/assert.kt
@@ -13,7 +13,45 @@ annotation class AssertkDsl
  * @see [assert]
  */
 @AssertkDsl
-class Assert<out T> internal constructor(val actual: T, val name: String?, internal val context: Any?) {
+sealed class Assert<out T>(val name: String?, internal val context: Any?) {
+    /**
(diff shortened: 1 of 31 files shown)


evant:
Thoughts about this in general and the compatibility story? Could be more aggressive here and break things immediately to not have to change twice in the long run.

evant:
Update: `BaseAssert` is now `Assert` which should ease migration issues. Instead, the `actual` property is deprecated an will immediately throw if it's not available.

Issue #160, opened 2018-12-28
Sort containsExactly output by index.


evant:
It would be much easier to compare output if matching indexes were next to each other in the output, instead of:

```
at index:0 expected: ...
at index:1 expected: ...
at index:0 unexpected: ...
at index:1 unexpected: ...
```

do

```
at index:0 expected: ...
at index:0 unexpected: ...
at index:1 expected: ...
at index:1 unexpected: ...
```

PR #161, opened 2019-01-04
containsExactly is now sorted by index

LinkedList:
Implements #160


evant:
Looks great, thanks for this!

PR #162, opened 2019-01-05
Rename assert to assertThat deprecating the old one

### LINKED ISSUE #146: Consider changing name of top level `assert` method to something that doesn't clash with Kotlin stdlib `assert`
I've had several teammates frequently run into confusion between assertk's `assert` and the Kotlin stdlib [`assert`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html). It has been confusing for some team members to see `import assertk.*`, have the imports folded in the editor, or when browsing the tests in the browser to have to cognitively differentiate the different call sites of `assertk`.

I thought I would open this to see what the opinions are around the naming.

evant:
Unfortuantly, assert clashes with kotlin's built-in assert methods which
cause import issues and potential confusion. By using the widely-used
assertThat instead, it's more clear what's using a flutent assert style.

Fixes #146

### DIFF
--- a/assertk-common/src/main/kotlin/assertk/assert.kt
+++ b/assertk-common/src/main/kotlin/assertk/assert.kt
@@ -10,7 +10,7 @@ annotation class AssertkDsl
 
 /**
  * An assertion. Holds an actual value to assertion on and an optional name.
- * @see [assert]
+ * @see [assertThat]
  */
 @AssertkDsl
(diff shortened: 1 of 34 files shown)


gvespucci:
Hi there! Is there a way to give this PR an opportunity to be approved? It looks like a real improvement to the library! :) 👍

gvespucci:
Dear maintainers, thank you very much for merging this PR! I really appreciated! :)

PR #164, opened 2019-01-13
Publish to maven central/github from tagged releases

evant:


evant:
Currently using my credentials, probably want to replace with WillowTree ones.

PR #173, opened 2019-02-13
Add Kotlin property (KProperty0) based assertThat method

maverick1601:


README.md
```diff
@@ -43,9 +43,15 @@ class PersonTest {
 
     @Test
     fun testAge() {
-        assertThat("age", person.age).isGreaterThan(20)
```
maverick1601: The order of arguments seemed to be wrong.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [0.13] - 2019-01-17
+## [0.13] - 2019-02-13
```
evant: 0.13 has already been released, please create a new [Unreleased] section for new changes.

README.md
evant: Good catch! Forgot to update the README when we changed the order.

assertk-common/src/main/kotlin/assertk/assert.kt
```diff
@@ -164,6 +165,20 @@ fun <T> assert(actual: T, name: String? = null): Assert<T> = assertThat(actual,
  */
 fun <T> assertThat(actual: T, name: String? = null): Assert<T> = ValueAssert(actual, name, null)
 
+/**
+ * Asserts on the given property value with an optional name.
```
evant: Should better explain why you'd use this version, particularly that it uses the property's name by default.

maverick1601: Changed the comment.

PR #175, opened 2019-02-20
Add isEqualToIgnoringGivenFields for JVM

Zomzog:
I've done it only for JVM because JavaScript Reflection can't list properties.


evant:
Sorry for the late response, I'm not sure I really understand the usecase for this? If you fell it's important you can open an issue describing it and we can discuss.

evant:
Hm, I guess it could be seen as the opposite of `isEqualToWithGivenProperties`, maybe with the name `isEqualToIgnoringGivenProperties`?

evant:
renamed and merged

PR #216, opened 2019-06-28
Added messageContains for Throwable

jarlehansen:
It is often useful to assert on the contents of an exception message without comparing the entire message. For instance it is useful to check the the message contains what the input was and not necessarily check the entire string.

I have added the function `fun Assert<Throwable>.messageContains(text: String)`. This is a feature I saw the need for in my own projects.


derektom14:
Perhaps you should have a method `Assert<Throwable>.message(): Assert<String>` method that allows you to use any string-related assertion on an error message, such as `startsWith` and getting ignore-case support.

evant:
We do! It's actually used in the implementation in this pr.

PR #218, opened 2019-07-06
Replace returnedValue, thrownError, and doesNotThrowAnyException…

evant:
… with `isSucess` and `isFailure`

This provides the following benefits:
- Makes them more inline with the chaining style of the rest of the
assertions (ex: `isSucess().isEqaulTo(1)` instead of
`returnedValue { isEqualTo(1) }`.
- Removes the last remaining instance of 'surprise' soft assertions. Now
you will _only_ get soft assertions using `assertAll` or `all`.
- Simpler implementation.

Note: Ideally this would be using assertions on the kotlin.Result type.
However there is a [bug](https://youtrack.jetbrains.com/issue/KT-32450)
with inline classes that prevents this. While we could wait for this to
be fixed before making this change, I opted instead to use our own
Result type in the meantime. I suspect that holding a reference to the
intermediate result type would not be very common so this should be a
mostly source-compatible change in the future. And where it's not, we
can introduce a `@Deprecated typealias assertk.Result<T> =
kotlin.Result<T>` to aid in migration.


geoffreywiseman:
Good changelog for these changes, BTW; thanks! 👍

PR #229, opened 2019-07-28
Add optional equivalent function for isEqualTo/isNotEqualTo

evant:
Sometimes you want to compare objects that don't have a proper equals
implementation. This makes it simpler to implement this. Ex:

```kotlin
fun Assert<CharSequence>.isEqualTo(other: CharSequence) {
    isEqualTo(other) { a, b -> a.toString() == b.toString() }
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ plugins {
 
 allprojects {
     group 'com.willowtreeapps.assertk'
-    version '0.19'
+    version '0.20-SNAPSHOT'
 }
 
 ext {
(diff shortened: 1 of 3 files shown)


Kritarie:
Consider also a separate `isEqualTo` where the types are the same and the caller can provide a property selector:
```fun Assert<Point>.isEqualTo(other: Point) {
    isEqualTo(other) { point -> point.x }
}```

evant:
https://truth.dev/api/latest/com/google/common/truth/Correspondence.html might be a good place to look for inspiration.

evant:
Ok changed it to add the new functions `corresponds`/`doesNotCorrespond`. Think this more clear then modifying `isEqualTo`. Chose to not do a property selector as we already have that with `isEqualToWithGivenProperties`.

PR #234, opened 2019-08-14
Add optional function to customize display of shown values

evant:


evant:
Closing, not very happy with the impl, can re-visit in the future if necessary.

PR #256, opened 2019-11-11
Add props to assert on keys or values of a map

evant:


src/commonTest/kotlin/test/assertk/assertions/MapTest.kt
```diff
@@ -175,4 +176,16 @@ class MapTest {
...
+    @Test fun extracts_size() {
+        assertEquals(1, assertThat(mapOf("one" to 1)).size().valueOrFail)
+    }
+
+    @Test fun extracts_keys() {
+        assertEquals(setOf("one", "two"), assertThat(mapOf("one" to 1, "two" to 2)).keys().valueOrFail)
+    }
+
+    @Test fun extracts_values() {
+        assertEquals(listOf(1, 2), assertThat(mapOf("one" to 1, "two" to 2)).values().valueOrFail.toList())
```
evant: `toList()` is needed because map uses a specialized collection for it's values and `AbstractCollection` does not have a default implementation of `equals()`

evant:
Closing, since we don't have a use-case for this.

PR #257, opened 2019-11-11
Render tabs and newlines in diffs to make it clear why they are different

### LINKED ISSUE #246: isEqualTo should report when two strings only differ by line endings
Similarly to [what Truth does](https://github.com/google/truth/blob/dba0b01d8690c6ee4c379fadc2ff28840a302961/core/src/main/java/com/google/common/truth/Platform.java#L128-L145), AssertK should report in the diff when the only differences between two strings which are supposed to be identical are in line breaks — this could save lots of people lots of WTFs :)

evant:
Fixes #246

### DIFF
--- a/src/commonMain/kotlin/assertk/assertions/support/support.kt
+++ b/src/commonMain/kotlin/assertk/assertions/support/support.kt
@@ -54,8 +54,10 @@ fun <T> Assert<T>.fail(expected: Any?, actual: Any?) {
         val extractor = DiffExtractor(display(expected), display(actual))
         val prefix = extractor.compactPrefix()
         val suffix = extractor.compactSuffix()
+        val expectedDiff = extractor.expectedDiff().renderSpecialWhitespace()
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/assertk/assertions/support/support.kt
```diff
@@ -86,3 +88,14 @@ private fun formatName(name: String?): String {
         " [$name]"
     }
 }
+
+private val specialWhitespace = Regex("[\r\n\t]|  +")
```
rock3r: Any specific reason for not using `\s`?

evant: Lack of familiarity with regex really.

evant: Actually, the space stuff in that regex is left over from when I was trying to test if they should be revealed (My original thought was to only reveal if the diff looked the same otherwise, ended up with just always showing them in the diff). Changing the pattern to just the chars that are being replaced.

PR #277, opened 2020-01-31
Add isEqualByComparingTo assertion methods

lukas-krecan:


evant:
Can you give a concrete example where you'd want to use this over `isEqualTo()`? Just trying to understand the user-case.

lukas-krecan:
The main use-case is comparing BigDecimals (as seen in the test) where 20.00 is not equal to 20 but it is equal when using comparator.

evant:
Hm, makes sense. This other option would be to have a special `isEqualTo()` for BigDecimal that using `compareTo` under the hood, like we do for arrays. Not sure if that's more confusing though.

lukas-krecan:
In my opinion, overloading `isEqualTo()` is indeed more confusing.

PR #278, opened 2020-02-02
Add multi-value support for String contains assertion

christophpickl:
hi, i am missing a few assertions in assertk, which for the time being reside in some privately hosted repository. i'd like to start sharing those with the community, as i think they are essential and should reside in the core project itself (rather get dusty in some private shared library).

sorry i haven't opened up an issue upfront to discuss this. i hope/guess it's simple enough to just go for it ;)

PS: this pull request was created due to the FOSDEM 2020 kotlin open source initiative (https://fosdem.org/2020/schedule/track/kotlin/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-- Add `Assert<String>.containsAll()` assertions
+- Add `Assert<String>.containsSubstrings()` assertions
 
 ## [0.21] 2020-01-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add `Assert<String>.containsAll()` assertions
+
 ## [0.21] 2020-01-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-- Add `Assert<String>.containsSubstrings()` assertions
+- Add multi-value support for `Assert<String>.contains()` and `doesNotContain()`
 
 ## [0.21] 2020-01-22
 
--- a/src/commonMain/kotlin/assertk/assertions/string.kt
+++ b/src/commonMain/kotlin/assertk/assertions/string.kt
@@ -1,6 +1,7 @@
 package assertk.assertions
 
 import assertk.Assert
+import assertk.assertAll
 import assertk.assertions.support.expected
 import assertk.assertions.support.fail
 import assertk.assertions.support.show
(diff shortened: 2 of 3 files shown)


src/commonMain/kotlin/assertk/assertions/string.kt
```diff
@@ -47,6 +48,26 @@ fun Assert<String>.contains(other: CharSequence, ignoreCase: Boolean = false) =
...
+}
+
+/**
+ * Asserts the string contains the expected strings.
+ * @param ignoreCase true to compare ignoring case, the default if false.
+ */
+fun Assert<String>.containsSubstrings(expected: List<String>, ignoreCase: Boolean = false) {
+    assertAll {
+        expected.forEach {
+            contains(it, ignoreCase)
```
christophpickl: this simply delegates to the already existing `Assert<String>.contains()`. another possibility would be to "inline" the logic like this:

```kotlin
fun Assert<String>.containsSubsrings(expected: List<String>, ignoreCase: Boolean = false) = given { actual ->
    expected.forEach {
        if (!actual.contains(it, ignoreCase)) {
            expected("to also contain:${show(it)} but was:${show(actual)}")
        }
    }
}
```

any ideas/opinions about that alternative?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add `Assert<String>.containsSubstrings()` assertions
```
christophpickl: i originally wanted to call it `containsAll`, but i've encountered some strange compilation errors doing so (some issues with ambiguous function calls i think)... 🤔

src/commonTest/kotlin/test/assertk/assertions/StringTest.kt
```diff
@@ -102,6 +102,35 @@ class StringTest {
...
+    //region containsSubstrings
+    @Test fun containsSubstrings_value_contains_passes() {
+        assertThat("test").containsSubstrings("es", "st")
+    }
+
+    @Test fun containsSubstrings_list_contains_passes() {
+        assertThat("test").containsSubstrings(listOf("es", "st"))
+    }
+
+    @Test fun containsSubstrings_value_not_contains_fails() {
```
evant: Could you add/replace these with ones that actually take multiple values?

src/commonMain/kotlin/assertk/assertions/string.kt
christophpickl: good point, error message. 

GIVEN:
```kotlin
assertThat("test").containsSubstrings("foo", "bar")
```

CURRENT OUTPUT:
```
The following assertions failed (2 failures)
	expected to contain:<"foo"> but was:<"test">
	expected to contain:<"bar"> but was:<"test">
```

changed/improved it to:
```
expected to contain:<["foo", "bar"]> but was:<"test">
```

CHANGELOG.md
christophpickl: done. so there is `contains(CharSequence)`, `contains(CharSequence...)` and `contains(Iterable<CharSequence>)` 👍 

(no need for `contains(Array<CharSequence>)`)

src/commonTest/kotlin/test/assertk/assertions/StringTest.kt
christophpickl: good point. even more necessary, now as the method is overloaded.

src/commonTest/kotlin/test/assertk/assertions/StringTest.kt
```diff
@@ -102,6 +102,35 @@ class StringTest {
...
+        val error = assertFails {
+            assertThat("Test").containsSubstrings("EST", ignoreCase = false)
+        }
+        assertEquals("expected to contain:<\"EST\"> but was:<\"Test\">", error.message)
+    }
+    //endregion
+
+
     //region doesNotContain
     @Test fun doesNotContain_value_not_substring_passes() {
```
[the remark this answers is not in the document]
christophpickl: good idea, consistency 👍 done.

PR #300, opened 2020-04-01
Make tests system-independent

derektom14:
Fixes tests to work on Windows, and be generally more system-independent:

- Use `File.separator` to be separator-agnostic
- Compare error messages using `lines()` to be line separator-agnostic

One change to the actual library behavior: `assert(file).hasPath(path)` originally would fail on Windows to consider that a file `File("assertKt/file.txt")` has a path `"assertKt/file.txt"` because the `/` is replaced with a `\`. So that the tests using assertk are also system-independent, the `hasPath` method is revised to compare the file to a file that also has the expected path.

### DIFF
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/file.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/file.kt
@@ -93,7 +93,7 @@ fun Assert<File>.hasName(expected: String) {
  * Asserts the file has the expected path.
  */
 fun Assert<File>.hasPath(expected: String) {
-    path().isEqualTo(expected)
+    path().isEqualTo(File(expected).path)
 }
 
 /**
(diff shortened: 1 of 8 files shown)


assertk/src/jvmMain/kotlin/assertk/assertions/file.kt
```diff
@@ -93,7 +93,7 @@ fun Assert<File>.hasName(expected: String) {
  * Asserts the file has the expected path.
  */
 fun Assert<File>.hasPath(expected: String) {
-    path().isEqualTo(expected)
+    path().isEqualTo(File(expected).path)
```
derektom14, line 96: This same change should probably also be applied to `hasParent`.

evant: Makes sense, can you go ahead and change that one as well? Would also recommend adding/updating the test so it would fail with the old way.

Issue #298, opened 2020-03-29
Double-clicking failing test name in Android Studio doesn't play well with `assertAll` 


tom-jb:
I realize this may be a problem in Android Studio (3.6.1) and not AssertK, but is there anything AssertK could do here?

**Reproducible steps:**

1. Create this file in a 'test' directory:

````kotlin
// Foo.kt
class Foo {
    @Test fun bar() {
        assertAll {
            assertThat(1).isEqualTo(2)
            assertThat(2).isEqualTo(3)
        }
    }
}
````

2. Run the tests.

3. Double-click the failing test function in the Run tool window.

Android Studio opens File.kt and positions the cursor at the *end of the file*.

Ideally, it would position the cursor at the first failing assertion. And if you remove the `assertAll` block from the above then, of course, it does exactly this.

**Comment by evant** [comment]
Yeah, that's probably because it's the `assertAll` block that ends up throwing the exception that's collected. To see if anything can be done about it, the best way would probably compare with junti5's `assertAll` to see if it's handled any differently.

**Comment by tom-jb** [comment]
Sorry for the late reply!

When I use JUnit 5, double-clicking takes me straight to the method name. The caret goes to the beginning of 'bar' below.

````kotlin
// Foo.kt
class Foo {
    @Test fun |bar() { // | is the caret
        assertAll({
            assertEquals(1, 2)
        }, {
            assertEquals(2, 3)
        })
    }
}
````

**Comment by evant** [comment]
Hm, it looks like it's because `assertAll` is marked inline, seems to confuse intellij here.

**Comment by evant** [comment]
I'm tempted to make `assertAll` not inline. The use-cases for using it that way are fairly limited imo.

**Comment by tom-jb** [comment]
Would that change performance much?

I ask because I think every assertion framework I've used other than yours has quite a slow `assertAll`-type system.

For example, as far as I remember, the equivalent in AssertJ added something like 100 or 200 milliseconds per invocation in some cases.

**Comment by evant** [comment]
I doubt it, but it would be interesting to benchmark.

**Comment by evant** [comment]
Results with jmh:

no assert-all:
```
Result "test.assertk.AssertBenchmarks.assertThat_isEqualTo":
  5.041 ±(99.9%) 0.064 ns/op [Average]
  (min, avg, max) = (4.956, 5.041, 5.210), stdev = 0.086
  CI (99.9%): [4.977, 5.106] (assumes normal distribution)
```

inline:
```
[...]

PR #301, opened 2020-04-28
Add flux assertion

Zomzog:
I think it can be useful to have assertions on coroutine flux directly.

I've tried it with JVM `containsExactly` and `index` but if you're ok with the proposal I can add others.


evant:
Yeah, have considered flow assertions, I think something would be a good idea. A few bits of feedback for your pr:
1. Because of the dependency I do think they should be in a separate module.
2. The assertions you added assume the flow terminates, which may not always be the case. Need to look into good ways on asserting on non-terminating flows. [rxjava's test assertions](http://reactivex.io/RxJava/javadoc/io/reactivex/observers/TestObserver.html) may be some good inspiration here.

evant:
Created my own pr https://github.com/willowtreeapps/assertk/pull/302 if you want to take a look. Some implementation decisions:

- Created a new artifact: `assertk-coroutines`
- The new flow assertions are all `suspend`. This allows you to use them on all supported platforms. In java, your likely to be wrapping your coroutine tests in a `runBlockingTest`, this plays nicely with that.
- `transform` has been made inline. This allows you to call suspend functions inside of it.
- Unstable coroutine api's are avoided for the most part. Except for cases where it's wrapping an unstable method: `count()`.
- Care is taken to support flows that don't complete when it makes sense.

Zomzog:
Ok I've block on `transform` and created a dirty `coTransform` :)

How do you want to continue? You merge your PR and I rebase/rework my assertions on your way?

evant:
Yep, that would work.

evant:
In terms of assertions to be implemented for flow, I'm skeptical about `index()`. Unlike a list, a flow re-starts anytime it's consumed so something like
```kotlin
assertThat(actual).all {
   index(0).isPositive()
   index(1).isNegative()
}
```
which works well for lists wouldn't work well for flow.

Zomzog:
For index with all yes, the flow will be partially consumed on each assertion. Flow has some notion of order so I'm not sure it is useless, but if you prefer I can remove it.

I've tried to allow using the `all` of your example. It forces me to expose some internal stuff. Maybe I'm doing it wrong?

evant:
Sorry for the delay, yeah I don't think a duplicate `all` function is a good idea. We could make it inline for it to work here, though we may run into https://github.com/willowtreeapps/assertk/issues/298. Will have to think about it. For now, your `containsExactly` looks good, if you want to focus this pr on just that I'll merge it in.

evant:
merged https://github.com/willowtreeapps/assertk/pull/321 which had containsExactly

PR #302, opened 2020-05-04
Started on addding coroutine assertions, particuarly flow

evant:
Some implementation decisions:

- Created a new artifact: `assertk-coroutines`
- The new flow assertions are all `suspend`. This allows you to use them on all supported platforms. In java, you are likely to be wrapping your coroutine tests in a `runBlockingTest`, this plays nicely with that.
- `transform` has been made inline. This allows you to call suspend functions inside of it.
- Unstable coroutine api's are avoided for the most part. Except for cases where it's wrapping an unstable method: `count()`.
- Care is taken to support flows that don't complete when it makes sense.

### DIFF
--- a/assertk-coroutines/build.gradle
+++ b/assertk-coroutines/build.gradle
@@ -11,7 +11,7 @@ plugins {
 
 allprojects {
     group 'com.willowtreeapps.assertk'
-    version '0.23-SNAPSHOT'
+    version '0.1-SNAPSHOT'
 }
 
 ext {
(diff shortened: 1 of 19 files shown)


assertk-coroutines/build.gradle
```diff
@@ -0,0 +1,336 @@
...
+
+task nativeTest {
+    dependsOn(linuxTest, macosTest, iosX64Test)
+}
+
+task test {
+    dependsOn(jvmTest, jsTest, nativeTest)
+}
+
+task detektMain(type: io.gitlab.arturbosch.detekt.Detekt) {
```
nishtahir, line 155: maybe import `io.gitlab.arturbosch.detekt.Detekt`

assertk-coroutines/build.gradle
```diff
@@ -0,0 +1,336 @@
...
+        js {}
+        jvm {}
+        linux {}
+        iosArm64 {}
+        iosX64 {}
+        macos {}
+    }
+}
+
+task dokkaCommon(type: org.jetbrains.dokka.gradle.DokkaTask) {
```
nishtahir, line 191: Should we import `DokkaTask`

assertk-coroutines/build.gradle
evant: Most of this was duplicated from the assertk build.gradle, I can do a follow-up pr to clean them up.

Issue #304, opened 2020-05-05
Assert<T>.prop(KCallable) is not compile-time safe


alllex:
Currently, the `prop` function is implemented the following way:

```kotlin
fun <T, P> Assert<T>.prop(callable: KCallable<P>) = prop(callable.name) {
    try {
        callable.call(it)
    } catch (e: InvocationTargetException) {
        // unwrap cause for a more helpful error message.
        throw e.cause!!
    }
}
```

In its docs it suggest the following usage:
```kotlin
assertThat(person).prop(Person::name).isEqualTo("Sue")
```

However, the signature of the function accepting `KCallable` is not compile-time safe, because there is no connection between the generic type parameters `T` and `P`.

As a consequence, if one mixes up the property references passed to the `prop` function, the error will occur at runtime. The following snippet, that compiles without errors or warnings, demonstrates the problem.

```kotlin
import assertk.assertThat
import assertk.assertions.isEqualTo
import assertk.assertions.prop

data class Person(val name: String)
data class NotPerson(val notName: String)

fun main() {
    val p = Person(name = "John")
    assertThat(p).prop(NotPerson::notName).isEqualTo("John")
}
```

Running the snipping produces the following result:
```
Exception in thread "main" java.lang.IllegalArgumentException: object is not an instance of declaring class
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at kotlin.reflect.jvm.internal.calls.CallerImpl$Method.callMethod(CallerImpl.kt:97)
	at kotlin.reflect.jvm.internal.calls.CallerImpl$Method$Instance.call(CallerImpl.kt:113)
	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:106)
	at kotlin.jvm.internal.CallableReference.call(CallableReference.java:138)
	at assertk.assertions.AnyJVMKt$prop$1.invoke(any.kt:80)
	at assertk.Assert.transform(assert.kt:31)
	at assertk.assertions.AnyKt.prop(any.kt:135)
	at assertk.assertions.AnyJVMKt.prop(any.kt:78)
[...]

PR #305, opened 2020-05-05
Deprecate prop with KCallable argument and add overload with KProperty1 argument

alllex:
Implements #304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `prop` function with `KProperty1` argument.
 
 ### Deprecated
-- Deprecated `prop` function with `KCallable` argument.
+- Deprecated `prop` function with `KCallable` argument. Use the new overload
+with type-safe `KProperty1` argument or another overload with explicit name and lambda.
 
 ## [0.22] 2020-03-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add `prop` function with `KProperty1` argument.
+
+### Deprecated
+- Deprecated `prop` function with `KCallable` argument.
+
 ## [0.22] 2020-03-11
 
 ### Added
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
@@ -91,26 +91,6 @@ fun <T, P> Assert<T>.prop(callable: KCallable<P>) = prop(callable.name) {
 }
 
 /**
- * Returns an assert that asserts on the given property.
- *
- * Example:
- * ```
- * assertThat(person).prop(Person::name).isEqualTo("Sue")
- * ```
- *
- * @param property Property on which to assert. The name of this
- * property will be shown in failure messages.
- */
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add `prop` function with `KProperty1` argument.
+
+### Deprecated
+- Deprecated `prop` function with `KCallable` argument.
```
evant, line 13: Nice job adding this! Would add the note about migrating to the version that takes an explicit name and lambda.

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -75,6 +75,7 @@ fun <T : Any> Assert<T>.isNotInstanceOf(jclass: Class<out T>) = given { actual -
  * assertThat(person).prop(Person::name).isEqualTo("Sue")
  * ```
  */
+@Deprecated("Use an overload with KProperty1 instead")
```
evant, line 78: Am I correct in thinking that existing code would automatically move to the `KProperty1` overload? If that's the case then anything still calling the deprecated method would not be a `KProperty1`. Would have the deprecation message point them to the `prop(name, extractor)` method instead.

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -100,8 +121,10 @@ fun <T : Any> Assert<T>.isDataClassEqualTo(expected: T) = given { actual ->
 private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?): Unit = given { actual ->
     if (actual == expected) return
     if (kclass != null && kclass.isData) {
-        for (prop in kclass.memberProperties) {
-            prop(prop).isDataClassEqualToImpl(prop.call(expected), prop.returnType.classifier as? KClass<*>)
+        for (memberProp in kclass.memberProperties) {
+            @Suppress("UNCHECKED_CAST")
+            val force = memberProp as KProperty1<T, Any?>
```
evant, line 126: Does this cast always succeed? I wonder why `kclass.memberProperties` returns `KCallable`.

CHANGELOG.md
alllex: Will do!

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
alllex: Agreed

alllex: It is a shame that dokka still does not support overload references 😔

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
alllex: This one is rather odd.
The `klass.memberProperties` returns `Collection<KProperty1<T, *>>`, which seems to be fine. But when you try to get an item from this collection, like here in the for-loop, the type of the item is inferred to `KProperty1<out Any, *>`. And this type does not suit the signature of `prop(KProperty1)` at all. That's why there is a cast.

I assume that the cast must always succeed for several reasons. First of all, it is the same cast that was already used in another more general function here: https://github.com/willowtreeapps/assertk/blob/master/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt#L125
Secondly, the generic parameter of the `kclass` always corresponds to the `T`, even though it is star-projected. This is because we start with the top-most data-class type in the user-facing `isDataClassEqualTo`. Then, whenever we go down into one of the properties, we use the property return type to keep the correspondence.
Last, but not least, the tests for the `isDataClassEqualTo`, which include nested data classes, still pass.

PR #309, opened 2020-08-14
Convert Assert<String> extensions into Assert<CharSequence> extensions.

Kritarie:
Most of the diff is a simple cut-paste from `strings.kt` and `StringTest.kt` to `charsequence.kt` and `CharSequence.kt`.
Slight modifications were required for `isEqualTo` and `isNotEqualTo`, as they both relied on the `String?` extension of `.equals` in the standard lib.
Also adds two tests that caught a bug I introduced with this change, specifically when comparing `null` to `"null"`

Couple thoughts:
1. Is there anything I'm missing that might cause a maintenance headache with this change?
2. Are there cases where this will break compilation for consumers?

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/charsequence.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/charsequence.kt
@@ -53,3 +53,94 @@ fun Assert<CharSequence>.hasSameLengthAs(other: CharSequence) = given { actual -
     if (actualLength == otherLength) return
     expected("to have same length as:${show(other)} ($otherLength) but was:${show(actual)} ($actualLength)")
 }
+
+/**
+ * Asserts the char sequence has the expected number of lines.
+ */
+fun Assert<CharSequence>.hasLineCount(lineCount: Int) = given { actual ->
+    val actualLineCount = actual.lines().size
+    if (actualLineCount == lineCount) return
+    expected("to have line count:${show(lineCount)} but was:${show(actualLineCount)}")
+}
+
+/**
+ * Asserts the char sequence contains the expected subsequence.
+ * @param ignoreCase true to compare ignoring case, the default if false.
+ */
+fun Assert<CharSequence>.contains(expected: CharSequence, ignoreCase: Boolean = false) = given { actual ->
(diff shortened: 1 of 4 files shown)


assertk/src/commonMain/kotlin/assertk/assertions/charsequence.kt
```diff
@@ -53,3 +54,122 @@ fun Assert<CharSequence>.hasSameLengthAs(other: CharSequence) = given { actual -
...
+
+/**
+ * Asserts the char sequence is equal to the expected char sequence.
+ * @param ignoreCase true to compare ignoring case, the default if false.
+ * @see [isNotEqualTo]
+ */
+fun Assert<CharSequence?>.isEqualTo(other: CharSequence?, ignoreCase: Boolean = false) = given { actual ->
+    val actualString = actual?.toString()
+    val otherString = other?.toString()
+    if (actualString.equals(otherString, ignoreCase)) return
```
evant, line 75: So it's possible for a `CharSequence` to not be equal but their `toString()` representations to be equal (Think spanned strings on android). Think we need to keep these only for `String`.

assertk/src/commonMain/kotlin/assertk/assertions/charsequence.kt
```diff
@@ -53,3 +54,122 @@ fun Assert<CharSequence>.hasSameLengthAs(other: CharSequence) = given { actual -
...
+    fail(other, actual)
+}
+
+/**
+ * Asserts the char sequence is not equal to the expected char sequence.
+ * @param ignoreCase true to compare ignoring case, the default if false.
+ * @see [isEqualTo]
+ */
+fun Assert<CharSequence?>.isNotEqualTo(other: CharSequence?, ignoreCase: Boolean = false) = given { actual ->
```
evant, line 84: Ditto for `isNotEqualTo`

assertk/src/commonMain/kotlin/assertk/assertions/charsequence.kt
Kritarie: Good point :+1:

PR #316, opened 2020-08-26
Adds containsExactlyInAnyOrder to iterable and primitive / float arrays.

Kritarie:


assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
```diff
@@ -73,6 +74,40 @@ fun Assert<Iterable<*>>.containsOnly(vararg elements: Any?) = given { actual ->
     }.toString())
 }
 
+/**
+ * Asserts the iterable contains exactly the expected elements, in any order.
```
evant, line 78: Think this doc could be more descriptive in describing the different between this and the other contains methods.

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
```diff
@@ -73,6 +74,40 @@ fun Assert<Iterable<*>>.containsOnly(vararg elements: Any?) = given { actual ->
...
+        if (notInActual.isNotEmpty()) {
+            append("\n elements not found:${show(notInActual)}")
+        }
+        if (notInExpected.isNotEmpty()) {
+            append("\n extra elements found:${show(notInExpected)}")
+        }
+    }.toString())
+}
+
+internal fun MutableList<*>.removeFirst(value: Any?) {
```
evant, line 106: Any reason this isn't private?

Kritarie: Nah, it should be private

PR #329, opened 2020-12-05
Add first and single assertions

derektom14:
This adds convenient `first` and `single` assertions.

I've seen this as a common use-case, particularly with small collections in tests. When the test is just equality, then usually `isEqualTo(listOf(singleItem))` will work, but if you want to make a different assertion, it becomes more complicated, usually having to split into `assertThat(list).hasSize(1)` and then `assertThat(list.first())....`.

The two test methods are modeled after the standard library's `Iterable.first()` and `Iterable.single()` methods.


exaV:
This is one of the few features I miss from assertJ (see https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractIterableAssert.html#first()) and it would be a great addition for assertk. 

@evant this PR has been open for a while. Is there anything I could help with to get the feature merged?

evant:
Squashed and merged

StefanLobbenmeier:
@evant are there any plans to release this as 0.26?

PR #335, opened 2021-01-16
Fix containsExactly on a custom list type

### LINKED ISSUE #332: containsExactly fails when using a custom List type.
Expected:

```kotlin
assertThat(MyCustomList("one", "two")).containsExactly("one", "two")
```

to pass, but gets

```kotlin
org.opentest4j.AssertionFailedError: expected to contain exactly:<["one", "two"]> but was:<["one", "two"]>
```

looks like `containsExactly` is using `==` to compare the lists, which a custom impl may not implement. It should iterate the elements instead.

note: found this using compose's `SnapshotStateList`.

evant:
Fixes #332


nishtahir (approved):
LGTM

kvithayathil (approved):

PR #353, opened 2021-04-27
Contains all in order #319

### LINKED ISSUE #319: Add `containsAllInOrder()`
Sometimes it's useful to assert a list contains a subset of items in order, but may have other items in the list.

andriesfc:
Closes #319

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/list.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/list.kt
@@ -57,17 +57,26 @@ private fun List<*>.contentEquals(other: Array<*>): Boolean {
 fun Assert<List<*>>.containsAllInOrder(sublist: List<*>) = given { actual: List<*> ->
 
     if (sublist.isEmpty() && actual.isEmpty()) return@given
-    val actualSublistStartIndex = if (sublist.isEmpty()) -1 else actual.indexOf(sublist.first())
-    val actualSubList = when {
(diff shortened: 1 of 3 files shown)


assertk/src/commonMain/kotlin/assertk/assertions/list.kt
```diff
@@ -46,4 +49,44 @@ private fun List<*>.contentEquals(other: Array<*>): Boolean {
...
+    val sublistMatchCount =
+        when (val firstMatchOfSublist = if (sublist.isEmpty()) -1 else actual.indexOf(sublist.first())) {
+            -1 -> 0
+            else -> {
+                var n = 1
+                if (n < sublist.size) do {
+                    val a = actual[firstMatchOfSublist + n]
+                    val b = sublist[n]
+                    n += 1
+                } while (a == b && n < sublist.size && n < actual.size)
```
evant, line 80: If feel like this would be more clear as a while loop, thoughts?

```suggestion
                while (n < sublist.size && n < actual.size) {
                    val a = actual[firstMatchOfSublist + n]
                    val b = sublist[n]
                    if (a != b) break
                    n += 1
                }
```

evant (changes requested):

assertk/src/commonMain/kotlin/assertk/assertions/list.kt
andriesfc: Thanx.  Much appreciated.  😄 👍

PR #356, opened 2021-04-29
Fixes #354 any {} fails with example from docs

### LINKED ISSUE #354: Another issue with nested any { it.all { } }
Following on from #331 (fixed now, thanks), another situation I've come across where the nested any, all doesn't behave as expected. Please check out my updated project: https://github.com/bschelberg/assertk-test-issue. Again there is just one test that should fail but it doesn't. Curiously, removing one of the pairs in the list results in the test failing as expected.

Test code:

    @Test
    fun `should not pass, pairs`() {
        val key1 = "key 1"
        val value1 = "value 1"

        val pairs = listOf("key 2" to "value 2", "key 2" to "value 2", "key 2" to "value 2", key1 to value1)

        assertThat(pairs).any {
            it.all {
                transform { it.first }.isEqualTo(key1)
                transform { it.second }.isEqualTo("unexpected")
            }
        }
    }

bschelberg:
The `+=` in the assertion meant that any {} only passed if one of the first 2 items in the Iterable passed

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
@@ -263,7 +263,7 @@ fun <E, T : Iterable<E>> Assert<T>.any(f: (Assert<E>) -> Unit) {
                     if (lastFailureCount == failure.count) {
                         itemPassed = true
                     }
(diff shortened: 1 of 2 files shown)


bschelberg:
Oh, I see you've already fixed this. I'll close this one.

PR #357, opened 2021-04-30
Add Path.exists assertion

yogurtearl:
gives a better error message and reads better than 
```kotlin
assertThat(Files.exists(regularFile)).isTrue()
```

### DIFF
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/path.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/path.kt
@@ -87,3 +87,15 @@ fun Assert<Path>.isSameFileAs(expected: Path) = given { actual ->
         expected("${show(actual)} to be the same file as ${show(actual)} but is not")
     }
 }
+
+/**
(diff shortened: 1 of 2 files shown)


yogurtearl:
@evant is there something I need to do to get the build to work for this PR? just showing as "Unauthorized" on Circle.

evant:
I don't know, will look at it when I get a chance. Likely no, you don't need to do anything.

evant (approved):

evant:
Maybe circle was having a day? I re-ran it and it worked. Seems like there's an error in the test though

> Task :assertk:compileTestKotlinJvm FAILED
e: /home/circleci/code/assertk/src/jvmTest/kotlin/test/assertk/assertions/PathTest.kt: (21, 29): Unresolved reference: of

evant:
Looks like it was because ci is running with java 8 and `Path.of` wasn't added until 11. The reason for this is because kotlin native was having issue on java 11 before. We may re-visit this but for now updated the test to use `Paths.get`.

PR #367, opened 2021-06-08
Added startsWith and endsWith assertions

ILikeYourHat:
Added `startsWith` and `endsWith` for `List` assertions.

This is my first contribution, so hello everybody 👋

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/list.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/list.kt
@@ -122,6 +122,8 @@ fun Assert<List<*>>.startsWith(vararg elements: Any?) = given { actual ->
             show(elements)
         }, but was:${
             show(sublist)
+        } in:${
+            show(actual)
         }"
     )
 }
@@ -152,6 +154,8 @@ fun Assert<List<*>>.endsWith(vararg elements: Any?) = given { actual ->
             show(elements)
         }, but was:${
             show(sublist)
+        } in:${
+            show(actual)
         }"
     )
 }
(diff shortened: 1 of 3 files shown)


assertk/src/commonMain/kotlin/assertk/assertions/list.kt
```diff
@@ -94,4 +94,64 @@ fun Assert<List<*>>.containsSubList(sublist: List<*>) = given { actual: List<*>
...
+    val sublist = if (actual.size >= elements.size) {
+        actual.subList(actual.size - elements.size, actual.size)
+    } else {
+        actual
+    }
+
+    if (sublist.contentEquals(elements)) return
+
+    expected(
+        "to end with:${
```
evant, line 151: Hm, the other assertions will show the entire list for context which I think is helpful. I think wouldn't be a problem to do that with `startsWith` but might be hard to read with `endsWith`, maybe showing the sublist in addition if necessary?

```kotlin
if (isSublist) {
  expected("to end with:${show(elements)} but was ${show(actual)}\n ending not expected:${show(sublist)}")
} else {
  expected("to end with:${show(elements)} but was ${show(actual)}")
}
```
you can see similar assertions in iterable.kt.

Thoughts?

evant (changes requested):

assertk/src/commonMain/kotlin/assertk/assertions/list.kt
ILikeYourHat: Maybe `expected("to end with:${show(elements)} but was ${show(sublist)} in ${show(actual)}` for both cases? This is some duplication, but you have sublist at the beginning to verify what's wrong, and full list at the end for context

evant: Sounds good

ILikeYourHat: changed

evant (approved):

PR #370, opened 2021-06-17
Add prop which takes Callable Function

Centronias:
This restores much of the functionality deprecated in 0.23, while preserving the useful type-safety of the new overloads.


Centronias:
I'm not sure if changes should be made to the deprecation message on the old `fun Assert.prop(KCallable)` variant.

evant:
Looks to be working well! I don't think it's worth bothering with the deprecated function, it should only show now when the types don't match so I don't think it's a big deal.

evant:
Rebased and merged

PR #372, opened 2021-07-04
Added support to assert against Java's Optional data type.

andriesfc:
Java's `Optional` type is used to indicate that method either has result, or none. 

A good example is a Spring Repository API which provides the following method: 

```java
Optional<T> findById(ID id)
```

This PR offers support for asserting against such an API.


README.md
```diff
@@ -188,6 +188,39 @@ tableOf("a", "b", "result")
 
 Up to 4 columns are supported.
 
+### Asserting against Java's `Optional` type
```
evant, line 191: Think adding this to the README is overkill, we don't include every assertion here as it would make things quite long. I think it being included in the javadoc is sufficient.

assertk/src/jvmMain/kotlin/assertk/assertions/optional.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * val empty = Optional.empty<Int>()
+ * assertThat(empty).value().isInstanceOf(Int::class) // Fails
+ * assertThat(empty).value().isEqualTo(0) // Fails
+ *
+ * ```
+ */
+fun <T> Assert<Optional<T>>.value(): Assert<T> {
+    return transform { actual ->
+        actual.orElseThrow {
```
evant, line 28: As much as I like using `orElseThrow` here, I think we should keep the messages consistent and use an `expected()` call instead.

evant (changes requested):
I like the idea, but would go with a bit different assertions to keep it consistent with other other assertions in this lib and the Optional api.

```kotlin
fun <T: Any> Assert<Optional<T>>.isPresent(): Assert<T>
```
and
```kotlin
fun Assert<Optional<*>>.isNotPresent()
```
This nicely mirrors the `isNull()` and `isNotNull()` assertions.

Additionally we could have a
```kotlin
fun <T: Any> Assert<Optional<T>>.hasValue(expected: T) = isPresent().isEqualTo(expected)
```
for the common case of matching an expected value much like the other `has` functions.

README.md
andriesfc: will remove it from the PR tonight after work. Thanx for the feedback

assertk/src/jvmMain/kotlin/assertk/assertions/optional.kt
andriesfc: Will change this. Thanx for the feedback. 👍🏼

evant:
Added some minor changes to update error messages/docs and test to better match existing code, other than that looked good!

PR #375, opened 2021-08-10
Skip private properties in isEqualToIgnoringGivenProperties

friscoMad:
The current code will try to get a property even if it is private and that throws an exception.

It is debatable if this function should work with classes with private fields as they can pass the check even if the private values are different.
In my use case, the private field is just a calculated value from the constructor params so it makes sense to skip it.

This change will not break any existing test as currently it is throwing on classes with private fields.


assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -130,7 +130,7 @@ private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?
 fun <T : Any> Assert<T>.isEqualToIgnoringGivenProperties(other: T, vararg properties: KProperty1<T, Any?>) {
     all {
         for (prop in other::class.members) {
-            if (prop is KProperty1<*, *> && !properties.contains(prop)) {
+            if (prop is KProperty1<*, *> && !properties.contains(prop) && prop.isAccessible) {
```
gvespucci, line 133: Do you think is it the case to add a proper test for this case, if not already provided? WDYT? Thanks

evant: Yes, an extra test on a class with a private var would be good.

evant:
Added test and merged

PR #377, opened 2021-08-20
Make index function generic

Bakrog:
Current list methods could be generalised with the use of collections. This pull request changes signature of `index` function but still compelling with projects that uses previous version.
This pull request should resolve #377


evant:
Hm, the reason for this is not all collections are ordered so it doesn't make sense to get the index. For example
`assertThat(HashSet(listOf(1, 2)).index(0)`
may return 1 or 2.

Bakrog:
This is exactly my use case for collections. I need to test if the result of my request is ordered so I use index to check it.

Bakrog:
@evant if this isn't a good way of how to test can you provide how you would do such test?

PR #382, opened 2021-09-08
New assertion for asserting multiple props of an object

mbjelac:
... with a readable specification syntax:

```kotlin
assertThat(MyObject("a", 1, true))
   .propsEqualTo(
     MyObject::name to "a",
     MyObject::amount to 1,
     MyObject::flag to true
   )
```


evant:
Hm, a little hesitant on having multiple ways on doing the same thing, since we have `prop()` and `isDataClassEqualTo()` but I don't hate it, will think about it.

mbjelac:
> Hm, a little hesitant on having multiple ways on doing the same thing, since we have `prop()` and `isDataClassEqualTo()` but I don't hate it, will think about it.

Yes, I've tried `prop` in this way but it can't be chained:

```kotlin
assertThat(something)
  .prop(Something::firstProp).isEqualTo(42)
  .prop(Something::secondProp).isEqualTo("foo")
```

and I don't see how it could be, since `isEqualTo` and others are "terminal" methods which execute the actual assertion.

`isDataClassEqualTo` asserts on all properties and often only some subset of them is important for a specific test.

the only existing tool that could be used is:

```kotlin
tableOf("propGetter", "expected")
  .row(Something::firstProp, 42)
  .row(Something::secondProp, "foo")
  .forAll { propGetter, expected ->
    assertThat(something)
      .prop(propGetter)
      .isEqualTo(expected)
  }
```

but its a bit too wordy and requires developers to remember that pattern of propGetter+expected.

evant:
You'd use `all` to chain `prop` fyi

```kotlin
assertThat(something).all {
  prop(Something::firstProp).isEqualTo(42)
  prop(Something::secondProp).isEqualTo("foo")
}
```

mbjelac:
> You'd use `all` to chain `prop` fyi
> 
> ```kotlin
> assertThat(something).all {
>   prop(Something::firstProp).isEqualTo(42)
>   prop(Something::secondProp).isEqualTo("foo")
> }
> ```

indeed ... here's a comparison from my code:

`propsEqualTo`:

```kotlin
assertThat(foundUser).propsEqualTo(
    FoundUser::role to UserRole.DOCTOR.roleName,
    FoundUser::firstName to firstName,
    FoundUser::lastName to lastName
)
```

vs. `all`

```kotlin
assertThat(foundUser).all {
    prop(FoundUser::role).isEqualTo(UserRole.DOCTOR.roleName)
    prop(FoundUser::firstName).isEqualTo(firstName)
    prop(FoundUser::lastName).isEqualTo(lastName)
}
```

`all` is a bit wordy-er however, it allows a mix of matchers (not forcing `isEqualTo` on every prop).

I'm currently inclined to drop this PR...

Issue #396, opened 2021-12-08
None Assertion for Iterables seems wrong


MatthiasGabriel:
Hi,

The none assertions for iterables does not work as I expect it.

The example in the assertion itself it the following:

```
    @Test
    internal fun `should be successful`() {
        assertThat(listOf("one", "two")).none {
            it.hasLength(2)
        }
    }
```

If I now adapt the example, by adding another item `ab` to the list, I'd expect the assertion to fail and the test case described below to be successful, however it fails.

```
    @Test
    public fun `should fail`() {
        assertThrows<AssertionFailedError> {
            assertThat(listOf("one", "two", "ab")).none {
                it.hasLength(2)
            }
        }
    }

```

In the initial pr https://github.com/willowtreeapps/assertk/pull/190 its also mentioned that it should behave as `noneMatch` in AssertJ

Therefore I compared the unit tests of your codebase against the behaviour of AssertJ.

One of the unit tests of the assertk code base:
```
    @Test fun each_non_matching_content_passes() {
        assertThat(listOf(1, 2, 3) as Iterable<Int>).none { it.isLessThan(2) }
    }
```

The unit test adapted to AssertJ, which throws an AssertionError, as expected by me.
```
   @Test
    void each_non_matching_content_passes() {
        assertThatThrownBy(() -> {
                Iterable<Integer> integers = newArrayList(1, 2, 3);
                assertThat(integers).noneMatch(s -> s < 2);
            }
        ).isInstanceOf(AssertionError.class);
    }
```

Thanks for having a look at it.

evant:
Should be fixed by https://github.com/willowtreeapps/assertk/pull/437

PR #397, opened 2021-12-08
Fix none assertion

MatthiasGabriel:
Fixes https://github.com/willowtreeapps/assertk/issues/396

Changes logic of none assertion, so it can be viewed as a "breaking" change

### DIFF
--- a/assertk/src/commonTest/kotlin/test/assertk/assertions/IterableTest.kt
+++ b/assertk/src/commonTest/kotlin/test/assertk/assertions/IterableTest.kt
@@ -3,7 +3,6 @@ package test.assertk.assertions
 import assertk.all
 import assertk.assertThat
 import assertk.assertions.*
-import assertk.assertions.support.fail
 import test.assertk.opentestPackageName
 import kotlin.test.Test
 import kotlin.test.assertEquals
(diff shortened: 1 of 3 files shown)


assertk/src/commonTest/kotlin/test/assertk/assertions/IterableTest.kt
```diff
@@ -224,8 +223,20 @@ class IterableTest {
...
 
+    @Test fun partial_matching_content_fails() {
+        val error = assertFails {
+            assertThat(listOf(1, 2, 3) as Iterable<Int>).none { it.isLessThan(2) }
+        }
+        assertEquals(
+            """expected none to pass (2 failures)
+                |${"\t"}${opentestPackageName}AssertionFailedError: expected [[1]] to be less than:<2> but was:<2> ([1, 2, 3])
+                |${"\t"}${opentestPackageName}AssertionFailedError: expected [[2]] to be less than:<2> but was:<3> ([1, 2, 3])
+               """.trimMargin().lines(), error.message!!.lines()
```
evant, line 234: Hm, error message is confusing here, we'd want the inverse right?

MatthiasGabriel: I agree on that. I'll have another look tomorrow to see how I can override the message and leave out the details.
I don't think there is a good way how to switch the message to contain the inverse message.

evant: Yeah, it'll probably require changes to how errors are collected to get the needed info.

MatthiasGabriel: Just wanted to shortly apologize, that I did not find time yet to have another look :)

evant: No worries! wasn't looking over the holidays anyway 😉

evant:

PR #404, opened 2022-01-28
Allow custom error message for boolean asserts

reisi007:
Signed-off-by: Florian Reisinger <reisi007@gmail.com>


evant:
I don't think this is a good way to add custom error messages as it doesn't generalize to all assertions, see https://github.com/willowtreeapps/assertk/discussions/352 & the linked issue for more discussion on this.

PR #405, opened 2022-02-07
Upgrade gradle version from 7.2 to 7.3.2

Sandhya2902:
Gradle upgrade to 7.3.2
https://docs.gradle.org/7.3.2/release-notes.html


evant:
Thanks for the pr, but I'd rather just wait until 7.4 is released is it comes with some improvements for version catalogs we can use.

PR #407, opened 2022-02-11
Upgrade Kotlin and related dependencies

pradeepblue:


evant:
What problem does this solve? We generally want to be conservative with kotlin version updates to make it more flexible for consumers.

pradeepblue:
> 
How badly does it affect consumer flexibility as it should be compatible for them if minimum version was set 1.5.30 earlier (4-5 months back)? Or, is the concern that other contributors might add code using some newer features? 
1.6.10 is stable and has improved optimizations (https://kotlinlang.org/docs/whatsnew16.html) that could be beneficial for the plugins used (gradle, dokka). If you still feel it is not yet ready for upgrade, I am ok with shelving it for few more months.

PR #409, opened 2022-03-17
Fix NoSuchElementException thrown by containsSubList

dzirbel:
Fix an edge case where calling containsSubList() throws a NoSuchElementException instead of an AssertionFailedError when the sublist is empty but the actual list is non-empty.

This was covered by unit tests, but they only verified that an exception was thrown, not the type of the exception. As such, I've specified the exception type via assertIs() which isn't used elsewhere in the codebase but seems appropriate to me. It's possible the general assertFails{} calls in unit tests should be replaced with a form that also checks the throwable type, to avoid this issue elsewhere.

Finally, it's my personal opinion that containsSubList() should _not_ fail for this case, i.e. accept [] as a sublist of ["one", "two"]. That matches the behavior of startsWith() and my intuition, but given that there was a unit test for this behavior I've opted to leave it as-is in this PR.


assertk/src/commonTest/kotlin/test/assertk/assertions/ListTest.kt
```diff
@@ -129,7 +131,8 @@ class ListTest {
     @Test fun containsSubList_fails_if_sublist_is_empty_and_actual_is_not_empty() {
         val emptySubList: List<String> = emptyList()
         val given: List<String> = listOf("Jason", "Jane", "Anne", "Darius", "Lee")
-        assertFails { assertThat(given).containsSubList(emptySubList) }
+        val throwable = assertFails { assertThat(given).containsSubList(emptySubList) }
+        assertIs<AssertionFailedError>(throwable)
```
evant, line 135: I think you can use [assertFailsWith](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-fails-with.html) to keep this as one line.

I would also check `AssertionError` instead of `AssertionFailedError`, as that should remain an implementation detail imo.

evant (commented):
>  It's possible the general assertFails{} calls in unit tests should be replaced with a form that also checks the throwable type, to avoid this issue elsewhere.

I think that's a good idea, feel free to submit a separate pr doing this, or I can when I have time.

> Finally, it's my personal opinion that containsSubList() should not fail for this case, i.e. accept [] as a sublist of ["one", "two"]. That matches the behavior of startsWith() and my intuition

I'm inclined to agree and would accept this behavior change.

assertk/src/commonTest/kotlin/test/assertk/assertions/ListTest.kt
dzirbel: Good call; will use assertFailsWith() to check for `AssertionError` in the future - thanks!

dzirbel:
> > It's possible the general assertFails{} calls in unit tests should be replaced with a form that also checks the throwable type, to avoid this issue elsewhere.
> 
> I think that's a good idea, feel free to submit a separate pr doing this, or I can when I have time.

Sounds good, I've submitted https://github.com/willowtreeapps/assertk/issues/410 to track. I'll probably take a stab at it today.

> > Finally, it's my personal opinion that containsSubList() should not fail for this case, i.e. accept [] as a sublist of ["one", "two"]. That matches the behavior of startsWith() and my intuition
> 
> I'm inclined to agree and would accept this behavior change.

Done, thanks!

PR #418, opened 2022-04-07
Add Troubleshooting section to README

hamen:
I faced a known issue that almost made me switch to Truth 😅
Eventually I found a solution, and I believe it would be useful to others as well.


hamen:
Unfortunately, I can't at the moment 😞

Issue #421, opened 2022-06-06
`exactly` works incorrectly with multiple assertions


rybick:
`exactly` assertion for collections works incorrectly when multiple assertions are provided in the assertion block.

Minimal failing scenario:
``` 
// snippet #1
    @Test
    fun test() {
        val c = listOf(1, -1, 3)
        assertThat(c).exactly(times = 1) {
            it.isNegative()
            it.isNegative()
        }
    }
```
Expected: `success`
got:
```
expected to pass exactly 1 times (4 failures)
	org.opentest4j.AssertionFailedError: expected [[0]] to be negative but was:<1> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[0]] to be negative but was:<1> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[2]] to be negative but was:<3> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[2]] to be negative but was:<3> ([1, -1, 3])
org.gradle.internal.exceptions.DefaultMultiCauseException: expected to pass exactly 1 times (4 failures)
	org.opentest4j.AssertionFailedError: expected [[0]] to be negative but was:<1> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[0]] to be negative but was:<1> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[2]] to be negative but was:<3> ([1, -1, 3])
	org.opentest4j.AssertionFailedError: expected [[2]] to be negative but was:<3> ([1, -1, 3])
```

The snippet below passes as expected:
```
// snippet #2
    @Test
    fun test() {
        val c = listOf(1, -1, 3)
        assertThat(c).exactly(times = 1) {
            it.isNegative()
        }
    }
```

Obvioulsy `snippet #1` is not a real world scenario, but multiple assertions in`exactly` are useful - for example if we had `isEven()` extension method:
```
assertThat(c).exactly(times = 1) {
    it.isNegative()
    it.isEven()
}
```
Or if we wanted to check just two fields out of many in an object:
```
assertThat(c).exactly(times = 1) {
    it.transform { it.country }
        .isEqualTo(UKRAINE)
    it.transform { it.city }
        .isEqualTo(KYIV)
}
```

rybick:
I've prepared a fix (https://github.com/willowtreeapps/assertk/pull/422), but it looks this repo is no longer mantained and there is no one to review it

[...]

PR #422, opened 2022-06-10
Fix `exactly` with multiple assertions (#421)

rybick:
Fix for a bug that occurred when `exactly` was called with multiple failing assertions described in #421


assertk/src/commonMain/kotlin/assertk/failure.kt
```diff
@@ -137,6 +142,12 @@ internal class SoftFailure(
     }
 }
 
+private class GroupedFailuresError(heading: String?, failures: List<Throwable>) : AssertionError() {
+    private val multipleFailuresError = MultipleFailuresError(heading, failures)
+
+    override val message: String? get() = multipleFailuresError.message
+}
```
rybick, line 149: I could inherit from `MultipleFailuresError` here, but that would require complicating `if` in the line `110`, so I believe this is cleaner.

rybick:
Anyone knows why the checks are stuck? @evant @rf43 @nishtahir

evant:

PR #423, opened 2022-06-21
use 'given' to wrap multiple assertions

katiedotson:
Fix for #421 uses `given` to wrap multiple assertions. One potential issue here is if there are multiple failures in a block, only the first will be shown as a cause. I might suggest making a separate function to solve for this.


rybick:
I have created a different fix for https://github.com/willowtreeapps/assertk/issues/421 a couple of days ago: https://github.com/willowtreeapps/assertk/pull/422
It doesn't have the problem that it shows only first error in a block, but is a little more complex.

evant:

PR #426, opened 2022-08-26
add list first()

mr-thierry:
Convenient method that allow a user to use `first()` instead of `index(0)`. This is similar to Kotlin `first()`.

I also took the opportunity to rephrase documentation of `index()`.


mr-thierry:
Closing because it's already in the code, just not in the current release

PR #439, opened 2023-03-16
Update to Kotlin 1.8.10 and all supported targets

JakeWharton:
Also:

- Remove explicit HMPP opt-in. It is now the default.
- Remove explicit JS script disabling. It is now the default.
- Check in yarn.lock which is now emitted and validated by default.
- Build and publish only on MacOS which can now cross-compile to Windows and Linux native targets.
- Only build for JS IR as the legacy compiler is deprecated for removal.
- Declare test dependency centrally as it now uses Gradle metadata to resolve to the correct platform-specific artifacts.
- Remove opt-in to opt-in which is stable as of Kotlin 1.7.


gradle/libs.versions.toml
```diff
@@ -1,11 +1,11 @@
...
-kotlin = "1.5.30"
+kotlin = "1.8.10"
 [libraries]
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
-dokka-gradle = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version = "1.5.0" }
+dokka-gradle = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version = "1.8.10" }
 detekt-gradle = { module = "io.gitlab.arturbosch.detekt:detekt-gradle-plugin", version = "1.19.0" }
-kotlin-coroutines = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.5.2" }
-opentest4k = { module = "com.willowtreeapps.opentest4k:opentest4k", version = "1.2.2" }
+kotlin-coroutines = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.7.0-Beta" }
```
JakeWharton, line 8: This beta is the first with all the targets needed. I would wait to release a new version until 1.7.0 final is out, unless you also want to do a beta.

evant: Yeah because of the low frequency of releases think it's best to wait

PR #445, opened 2023-03-16
Add unsigned arrays to array template

JakeWharton:
These are still experimental and so an annotation propagating that is added to the codegen for these types only.

I've modified the `TemplateTask` to support multiline declarations. I did it in a separate commit since it's technically orthogonal to the unsigned change.


evant (approved):

JakeWharton:
Good to merge?

evant:
Screenshot_20230319-122954_Firefox (1)
😕

JakeWharton:
Ah it only tells me this:



I'll get it ready for a rebase instead tonight.

JakeWharton:
Better?

PR #447, opened 2023-03-20
Configure languageVersion via toolchain

Goooler:
Follow up #439.

https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support


JakeWharton:
I believe this forces the installation of JDK 8 whereas mine will cross-compile using any available JDK.

Goooler:
Yeah, I just noticed only JDK 16 was used on CI, we can add more JDKs if necessary.

JakeWharton:
I would prefer not to land something like this, as the host JDK version should not matter since they can all cross-compile. We just need an appropriate minimum.

Goooler:
Alright.

PR #443, opened 2023-03-16
Upgrade to Gradle 8.0.2

JakeWharton:
Version catalogs are now stable and do not require enabling.

PR #451, opened 2023-04-18
Add implicit task dependency workaround

JakeWharton:
This works around a Kotlin multiplatform publishing bug.

See https://youtrack.jetbrains.com/issue/KT-46466

Refs #443


evant (approved):

JakeWharton:
Woo it worked!

Issue #453, opened 2023-04-24
New entrypoint: `assertFailure`


JakeWharton:
I'm finally getting around to migrating a bunch of assertions and I find this to be a frequent enough pattern to warrant its own entrypoint.

First, consider the very common pattern `assertThat(foo.bar()).isEqualTo("bar")`. This line is implicitly asserting that `foo.bar()` is successful and that the return value is `"bar"`.

Now another way to write this is:
```kotlin
assertThat { foo.bar() }
  .isSuccess()
  .isEqualTo("bar")
```
But that's so verbose! I _always_ expect the function call to succeed. I just want to assert on the value.

Now invert the problem: I want to test a function which I expect to _always_ fail.
```kotlin
assertThat { foo.bar() }
  .isFailure()
  .hasMessage("You're not 21!")
```

Even worse, in the case of functions that fail they usually fail in with particular exception types.
```kotlin
assertThat { foo.bar() }
  .isFailure()
  .isInstanceOf(UnderageException::classs)
  .hasMessage("You're not 21!")
```

The kotlin.test APIs have `assertFails` and `assertFailsWith` functions for these two cases. These are entrypoints where you are invoking a function whose result is failure and whose result is failure of a particular type.
```kotlin
assertFails { foo.bar() } // Assert<Throwable>
  .hasMessage("You're not 21!")
```
```kotlin
assertFailsWith<UnderageException> { foo.bar() } // Assert<UnderageException>
  .hasMessage("You're not 21!")
```

I believe kotlin.test also supports supplying a string message to match on, but in my opinion that's going too far and that's where the normal assertion APIs thrive (not to mention validating causes, other properties, etc.).

What do you think?

evant:
Hm, what about instead of an additional entrypoint we add convenience assertions on results?

```kotlin
assertThat { foo.bar() }
   .isFailureWith(UnderageException::class)
```

```kotlin
assertThat { foo.bar() }
   .isFailureWithMessage("You're not 21!")
```

evant:
Yeah that does make some sense. My specific concern with `assertFailsWith<UnderageException>` is that it moves the assertion itself to the entrypoint which is different from every other kind of assertion and backwards from the 'normal' api. This still somewhat happens with `assertFails` too, unlike `assertThat` it actually _does_ something by itself.  Maybe that's fine, though a common problem I've seen is people doing `assertThat` without an assertion on it and I wonder if this'll add more to that confusion.

[...]

PR #457, opened 2023-04-27
Add 'containsMatch' assertion for CharSequence

### LINKED ISSUE #452: New assertion: `containsMatch`
A mixture of `contains` and `match`, but only looks for a match within the string rather than against the whole thing.

Implementation is basically `matchesPredicate(regex::containsMatchIn)`, but it's nice to not have to do that every time you want to switch from an exact `contains` to a fuzzy, regex-based one (or similarly, from matching the whole string to only part of a string).

JakeWharton:
Unlike 'contains' this is a regex rather than a literal. Unlike 'matches' this looks for a subset of the charsequence to match rather than the entire contents.

Closes #452

### DIFF
--- a/assertk/src/commonTest/kotlin/test/assertk/assertions/CharSequenceTest.kt
+++ b/assertk/src/commonTest/kotlin/test/assertk/assertions/CharSequenceTest.kt
@@ -270,4 +270,18 @@ class CharSequenceTest {
         assertEquals("expected to match:${show(regex)} but was:<\"12345\">", error.message)
     }
     //endregion
+
+    //region contains match
+    @Test fun contains_match_matching_value_passes() {
+        assertThat("abc1234xyz").containsMatch(Regex("\\d\\d\\d\\d"))
+    }
+
+    @Test fun contains_match_not_matching_value_fails() {
+        val regex = Regex("\\d\\d\\d\\d")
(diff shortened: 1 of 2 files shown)


assertk/src/commonTest/kotlin/test/assertk/assertions/CharSequenceTest.kt
```diff
@@ -270,4 +270,18 @@ class CharSequenceTest {
...
+    //region contains match
+    @Test fun contains_match_matching_value_passes() {
+        assertThat("abc1234xyz").containsMatch(Regex("\\d\\d\\d\\d"))
+    }
+
+    @Test fun contains_match_not_matching_value_fails() {
+        val regex = Regex("\\d\\d\\d\\d")
+        val error = assertFailsWith<AssertionError> {
+            assertThat("abc123xyz").containsMatch(regex)
+        }
```
JakeWharton, line 283: It's a fun data point for #453 that the tests within the library use `assertFailsWith`!

evant: heh, that's because I didn't want to write the assertions with itself as that felt a little circular...

evant (approved):

PR #459, opened 2023-04-27
Simplify generics of `isInstanceOf`

JakeWharton:
`S` is only used to limit the receiver which means we can replace it with its bound (`Any`) and let the `out` variance support any subtype (such as `Assert<String>`) remain valid.

As far as I can tell this second generic was introduced as part of b817dc3c8b7915b3bbb147a4d82144dd377a478a, where `S` is actually used to hold an instance. But when the lambda was eventually removed the extra generic stuck around until today.


JakeWharton:
It's also just generally not true that the only instances you'd want to check are subtypes of the subject type. I can have a List and want to check that it implements the RandomAccess marker interface, for example. Those types don't share a common ancestor (besides Any).

evant (approved):

PR #460, opened 2023-04-27
Attempt to propagate a source throwable as a cause

### LINKED ISSUE #456: Consider adding actual `Throwable` as cause of `AssertionFailedError`
If I do
```kotlin
assertThat { somethingWhichFails() }
  .isFailure()
  .isInstanceOf(IllegalStateException::class)
```
and something _else_ fails throwing a different error type, all I see is:
```
expected to be instance of:<class java.lang.IllegalArgumentException> but had class:<class java.lang.AssertionError> (Failure(java.lang.AssertionError))
```
That isn't helpful!

It's a special casing, but in the case of the "actual" being a `Throwable` or a `Result` that's a failure, I think the `fail` function should add that `Throwable` as the cause to the `AssertionFailedError`.

This would provide me with the complete stacktrace of the unexpected failure and not just that the type or message disagrees.

We do this in Turbine and it's _excellent_. I would like it here, too!

JakeWharton:
This ensures that the full stacktrace can be seen to help identify the underlying source of a mismatch.

Consider the following two tests:
```kotlin
@Test fun demo1() {
  assertThat { someFunction() }
    .isFailure()
    .isInstanceOf(IllegalStateException::class)
}
@Test fun demo2() {
  assertThat { someFunction() }
    .isFailure()
    .hasMessage("failure!")
}
fun someFunction() {
  deepFunction()
}
fun deepFunction() {
  throw AssertionError("oh no!!!!!")
}
```
Demo 1's failure before:

    org.opentest4j.AssertionFailedError: expected to be instance of:<class java.lang.IllegalStateException> but had class:<class java.lang.AssertionError> (Failure(java.lang.AssertionError: oh no!!!!!))
    	at app//test.assertk.FailureTest.demo1(FailureTest.kt:56)
    	... 43 more

Demo 1's failure after:

    org.opentest4j.AssertionFailedError: expected to be instance of:<class java.lang.IllegalStateException> but had class:<class java.lang.AssertionError> (Failure(java.lang.AssertionError: oh no!!!!!))
    	at app//test.assertk.FailureTest.demo1(FailureTest.kt:56)
    	... 43 more
    Caused by: java.lang.AssertionError: oh no!!!!!
    	at test.assertk.FailureTest.deepFunction(FailureTest.kt:70)
    	at test.assertk.FailureTest.someFunction(FailureTest.kt:66)
    	at test.assertk.FailureTest.demo1(FailureTest.kt:54)
    	... 43 more

Demo 2's failure before:

    org.opentest4j.AssertionFailedError: expected [message]:<"[failure]!"> but was:<"[oh no!!!!]!"> (Failure(java.lang.AssertionError: oh no!!!!!))
    	at app//test.assertk.FailureTest.demo2(FailureTest.kt:61)
    	... 43 more

Demo 2's failure after:

    org.opentest4j.AssertionFailedError: expected [message]:<"[failure]!"> but was:<"[oh no!!!!]!"> (Failure(java.lang.AssertionError: oh no!!!!!))
    	at app//test.assertk.FailureTest.demo2(FailureTest.kt:62)
    	... 43 more
    Caused by: java.lang.AssertionError: oh no!!!!!
    	at test.assertk.FailureTest.deepFunction(FailureTest.kt:70)
    	at test.assertk.FailureTest.someFunction(FailureTest.kt:66)
    	at test.assertk.FailureTest.demo2(FailureTest.kt:60)
    	... 43 more

With the causes attached it's much easier to locate the source of the underlying problem.

Closes #456

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt
@@ -2,6 +2,7 @@ package assertk.assertions.support
 
 import assertk.Assert
 import assertk.NONE
+import assertk.ValueAssert
 import assertk.fail
 
 /**
@@ -76,10 +77,30 @@ fun <T> Assert<T>.fail(expected: Any?, actual: Any?) {
 fun <T> Assert<T>.expected(message: String, expected: Any? = NONE, actual: Any? = NONE): Nothing {
     val maybeSpace = if (message.startsWith(":")) "" else " "
     val maybeInstance = if (context.originatingSubject != null) " (${context.displayOriginatingSubject()})" else ""
+
+    // Attempt to extract a helpful Throwable to use as a cause if the current value
+    // or originating subject are a Throwable or failure Result.
+    var cause: Throwable? = null
+    if (this is ValueAssert<*>) {
+        cause = when (value) {
+            is Throwable -> value
+            is Result<*> -> value.exceptionOrNull()
+            else -> null
+        }
+        if (cause == null) {
+            cause = when (val originating = context.originatingSubject) {
+                is Throwable -> originating
+                is Result<*> -> originating.exceptionOrNull()
+                else -> null
+            }
+        }
+    }
+
     fail(
         message = "expected${formatName(name)}$maybeSpace$message$maybeInstance",
         expected = expected,
-        actual = actual
+        actual = actual,
+        cause = cause,
     )
 }
 
--- a/assertk/src/commonMain/kotlin/assertk/failure.kt
+++ b/assertk/src/commonMain/kotlin/assertk/failure.kt
@@ -4,6 +4,7 @@ import assertk.Failure.Companion.soft
 import com.willowtreeapps.opentest4k.AssertionFailedError
 import com.willowtreeapps.opentest4k.MultipleFailuresError
 import com.willowtreeapps.opentest4k.failures
+import kotlin.DeprecationLevel.HIDDEN
 
 /**
  * Assertions are run in a failure context which captures failures to report them.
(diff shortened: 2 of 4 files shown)


assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt
```diff
@@ -76,10 +77,30 @@ fun <T> Assert<T>.fail(expected: Any?, actual: Any?) {
...
+            else -> null
+        }
+        if (cause == null) {
+            cause = when (val originating = context.originatingSubject) {
+                is Throwable -> originating
+                is Result<*> -> originating.exceptionOrNull()
+                else -> null
+            }
+        }
+    }
```
JakeWharton, line 97: I started off with a bunch of clever functional shit here and it was an unreadable mess.

evant: mood

assertk/src/commonMain/kotlin/assertk/failure.kt
```diff
@@ -139,17 +140,29 @@ fun fail(error: AssertionError): Nothing {
 
 internal val NONE: Any = Any()
 
+// TODO Delete this before 1.0.
+@Deprecated("For binary compatibility", level = HIDDEN)
+fun fail(message: String, expected: Any? = NONE, actual: Any? = NONE): Nothing {
```
evant, line 145: I really haven't been bothering with binary compat before 1.0 (as opposed to source compat) with the assumption being that assertk is likely to be a leaf dependency instead of a transitive one. Would be happy to hear your thoughts though.

evant (commented):
Overall I like it, the added context does seem helpful

assertk/src/commonMain/kotlin/assertk/failure.kt
JakeWharton: It seemed like extremely low effort to retain so I included it. I could see a published library of assertions existing which might have used the function. That being said, if you don't want to set a precedent or anything I'm not tied to it.

evant: yeah it doesn't hurt anything, not sure I want to make binary compat a requirement for updates but I won't block merging this in on that.

evant (approved):

PR #461, opened 2023-04-27
Add `assertFailure` entrypoint

### LINKED ISSUE #453: New entrypoint: `assertFailure`

JakeWharton:
This accepts a lambda which is expected to throw an exception which is wrapped into an `Assert<Throwable>`. Successful completion of the lambda block will result in a failure exception.

Closes #453. Pairs well with #460.

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -184,12 +184,22 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
  *   throw Exception("error")
  * }.isFailure().hasMessage("error")
  * ```
+ *
+ * @see assertFailure
  */
-inline fun <T> assertThat(f: () -> T): Assert<Result<T>> = assertThat(Result.runCatching { f() })
+inline fun <T> assertThat(f: () -> T): Assert<Result<T>> = assertThat(runCatching(f))
 
 /**
  * Runs all assertions in the given lambda and reports any failures.
  */
 inline fun assertAll(f: () -> Unit) {
     Failure.soft().run { f() }
+}
+
+/**
+ * Asserts that the given block will throw an exception rather than complete successfully.
+ */
+fun assertFailure(f: () -> Unit): Assert<Throwable> {
+    runCatching(f).onFailure { return assertThat(it) }
+    fail("expected failure but lambda completed successfully")
 }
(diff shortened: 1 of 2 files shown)


JakeWharton:
Ah, yes. Will fix tomorrow.

assertk/src/commonMain/kotlin/assertk/assert.kt
```diff
@@ -184,12 +184,22 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
...
     Failure.soft().run { f() }
-}
\ No newline at end of file
+}
+
+/**
+ * Asserts that the given block will throw an exception rather than complete successfully.
+ */
+fun assertFailure(f: () -> Unit): Assert<Throwable> {
+    runCatching(f).onFailure { return assertThat(it) }
```
JakeWharton, line 203: Started with a simple try/catch, which I prefer, but it offended Detekt and I don't know how to suppress it. So instead I outsmarted it. `runCatching`/`Result` allocates on the exception path which annoys me because we immediately unbox it, but it's not like it matters in practice.

evant: You can normally suppress with `@Suppress("RULE NAME")` on the function. Detekt has a few code style rules that I think are too picky and I would like to get rid of, just haven't gotten around to it.

evant (approved):

PR #462, opened 2023-04-27
Make tests 20s faster

JakeWharton:
Any non-zero value will trigger suspension. Let's do 10 milliseconds instead of 10 seconds to save 20 seconds from wall clock time.


evant:
Lol good catch. It's because it was originally written with `runBlockingTest` which would skip the delay. Then it was updated to `runBlocking` and this was missed.

evant (approved):

PR #465, opened 2023-05-09
Remove lambda-accepting `assertThat` overload

### LINKED ISSUE #464: Consider eliminating `assertThat(body: () -> T)` entrypoint
With #461 on track to land (🤞), I don't think the lambda-accepting `assertThat` holds its weight anymore.

Expressions whose evaluation is expected to succeed can flow directly into `assertThat` to get wrapped into an `Assert<T>`. Either:
```kotlin
assertThat(foo()).…
```
or
```kotlin
val foo = foo()
assertThat(foo).…
```

Expressions whose evaluation is expected to fail can be enclosed within `assertFailure` to get wrapped into an `Assert<Throwable>`.
```kotlin
assertFailure { foo() }.…
```

Both of these handle the unexpected case.

`assertThat(foo())` with a throwing `foo()` simply sees the failure propagate to fail the test directly without ever entering this library.

`assertFailure { foo() }` with a `foo()` that completes successfully results in an `AssertionError` from `assertFailure`.

What about `Result<T>`? I think the only reason you would see a `Result<T>` now is if it was returned directly to you, and the library still has narrowing assertions extensions on `Assert<Result<T>>` to handle that case.
```kotlin
fun trySomething(): Result<String> { .. }

assertThat(trySomething()).isSuccess().isEqualTo("success!")
// or
assertThat(trySometing()).isFailure().isInstanceOf<EOFException>()
```

There's no need to use `assertThat { .. }` anywhere.

However, I have one final argument against it and it's the strongest one:

https://user-images.githubusercontent.com/66577/236091121-81f9fde3-1ec7-48ac-944e-59505c9a05a7.mov

It auto-completes ahead of `assertThat(T)` and that's basically never what you want!

JakeWharton:
Use `assertFailure` for always-failing things and `assertThat(value)` for always-successful or `Result`-returning things.

Closes #464

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -187,7 +187,7 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
  *
  * @see assertFailure
  */
-@Deprecated("Use assertThat or assertFailure", ReplaceWith("assertThat(runCatching(f))"))
+@Deprecated("Use assertThat(result) or assertFailure", ReplaceWith("assertThat(runCatching(f))"))
 inline fun <T> assertThat(f: () -> T): Assert<Result<T>> = assertThat(runCatching(f))
 
 /**
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -187,6 +187,7 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
  *
  * @see assertFailure
  */
+@Deprecated("Use assertThat or assertFailure", ReplaceWith("assertThat(runCatching(f))"))
 inline fun <T> assertThat(f: () -> T): Assert<Result<T>> = assertThat(runCatching(f))
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Changed
+- Lambda-accepting `assertThat { }` entrypoint is now deprecated.
+  Use `assertThat(T)` for normal values or `assertFailure { }` for exception-throwing code.
 - Previous deprecations are now errors
 
 ### Added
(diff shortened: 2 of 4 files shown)


JakeWharton:
I'll do a changelog pass in a follow-up with all the recent stuff since I've been slacking.

assertk/src/commonMain/kotlin/assertk/assert.kt
```diff
@@ -174,21 +174,6 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
...
- * assertThat { 1 + 1 }.isSuccess().isPositive()
- *
- * assertThat {
- *   throw Exception("error")
- * }.isFailure().hasMessage("error")
- * ```
- *
- * @see assertFailure
- */
-inline fun <T> assertThat(f: () -> T): Assert<Result<T>> = assertThat(runCatching(f))
```
evant, line 190: I know that it doesn't help the ide annoyance but I think this should go through one round of deprecation before removal. Would also allow `ReplaceWith` for easier updates.

JakeWharton: Updated.

Kotlin has an internal `@LowPriorityInOverload` annotation, but sadly it seems to only be available for the JVM and JS targets so it wouldn't help common code, sadly.

assertk/src/commonMain/kotlin/assertk/assert.kt
```diff
@@ -187,6 +187,7 @@ fun <T> Assert<T>.all(body: Assert<T>.() -> Unit) {
  *
  * @see assertFailure
  */
+@Deprecated("Use assertThat or assertFailure", ReplaceWith("assertThat(runCatching(f))"))
```
evant, line 190: A warning saying to use assertThat instead assertThat seems confusing? Maybe
```suggestion
@Deprecated("Use assertThat(result) or assertFailure", ReplaceWith("assertThat(runCatching(f))"))
```

evant (approved):

assertk/src/commonMain/kotlin/assertk/assert.kt
JakeWharton: Deprecation already fixes the sorting

Screen Shot 2023-05-12 at 9 10 27 PM

PR #472, opened 2023-05-15
Make assertFailure inline

### LINKED ISSUE #471: assertFailure isn't inline
We have several usages of `assertThat { ... }` that test suspending functions. This works well with `assertThat` since it's inline but the new replacement, `assertFailure`, is not inlined so it cannot be used to test suspending functions.

ansman:
This change makes `assertFailure` inline so it can be used to test suspending functions.

This fixes #471

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -200,7 +200,7 @@ inline fun assertAll(f: () -> Unit) {
 /**
  * Asserts that the given block will throw an exception rather than complete successfully.
  */
-fun assertFailure(f: () -> Unit): Assert<Throwable> {
+inline fun assertFailure(f: () -> Unit): Assert<Throwable> {
     @Suppress("TooGenericExceptionCaught") // Intentionally capturing all exceptions.
     try {
         f()


assertk/src/jvmTest/kotlin/test/assertk/JVMAssertFailureTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertSame
+import kotlin.test.assertTrue
+
+class JVMAssertFailureTest {
+    @Test
+    fun can_test_suspending_functions(): Unit = runBlocking {
```
JakeWharton, line 21: This could be in `commonTest` by using `runTest`. I was going to migrate the `assertThat { }` tests to use it, but then with its deprecation it seemed pointless.

If you want I can migrate it in a follow-up.

ansman: Ah, I forgot they had changed it from `runBlockingTest` and I assumed it wasn't available for common. Will fix.

ansman: Done

evant (approved):

PR #478, opened 2023-06-26
Do not allow null-based assertions on non-null types

JakeWharton:
Before:
Screen Shot 2023-06-26 at 10 23 12 AM

After:
Screen Shot 2023-06-26 at 10 23 23 AM


evant:
Clever! This doesn't break assertions on platform types right? I don't think it would?

JakeWharton:
Sadly it selects the non-null variant :(

JakeWharton:
I'm assuming that is a deal-breaker. I can't see a path forward. It's weird that it selects the non-null variant. I spent a little time trying to investigate ways to trick it and couldn't come up with anything.

evant:
Yeah you'd need to assert both on a platform type. Could do something weird like only added them to the native variant? Not sure it's worth though.

Issue #479, opened 2023-07-10
WASM support with Kotlin 1.9


JakeWharton:
1. [x] https://github.com/willowtreeapps/opentest4k/pull/26
2. [x] OpenTest4k release
3. [ ] AssertK Kotlin 1.9 + WASM PR

evant:
opentest4k 1.2.4 released

JakeWharton:
Thanks! I will proceed with a PR here soon.

PR #481, opened 2023-07-24
Update to Kotlin 1.9

JakeWharton:
Refs #479. Actual WASM support will come in a follow-up as it's a little more complicated than I hoped (due to usage of coroutines in main tests, and lack of WASM coroutines).


eygraber:
@JakeWharton is this sufficient for the coroutines - https://maven.pkg.jetbrains.space/kotlin/p/wasm/experimental/org/jetbrains/kotlinx/kotlinx-coroutines-test-wasm/

evant (approved):

PR #482, opened 2023-08-10
Fix the build on Windows

carltonwhitehead:
The project had been failing to build on my Windows PC with the following rather cryptic exception:
Caused by: java.lang.StringIndexOutOfBoundsException: begin 324, end -1, length 5122

I tracked this down to use of hard-coded Unix-style line separator "\n" in TemplateTask.kt. Using the system's line separator corrected the problem.

I don't think this is likely to have any impact on non-Windows builds, as these templates are evidently dependent on the build platform, but releases already work on Windows, and clearly the build infrastructure is not Windows.

### DIFF
--- a/buildSrc/src/main/kotlin/TemplateTask.kt
+++ b/buildSrc/src/main/kotlin/TemplateTask.kt
@@ -61,6 +61,7 @@ abstract class TemplateTask : DefaultTask() {
 
     @TaskAction
     fun run() {
+        val lineSeparator = System.lineSeparator()
         val inDir = inputDir.get().asFile
         val outDir = outputDir.get().asFile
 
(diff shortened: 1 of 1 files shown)


carltonwhitehead:
I'm not sure anyone intended this project to build on Windows. I had cloned it to contribute back a few custom assertions, and noticed I couldn't get the project to build, so I went ahead and tried to fix it.

It might be worth also adding a Windows executor to the CircleCI to help verify this change and protect against regressions. I'd be open to working on that too.

fmedlin:
Good catch! LGTM

JakeWharton (approved):
Ah this was probably my fault. Good catch.

evant (approved):

evant:
@carltonwhitehead  Yeah, would have no problem adding a windows executor. Right now the macos executor builds the mac binaries and the linux executor builds the rest. So there's actually a couple of ways we could go with this.

1. Just add the windows executor to test it builds but don't change anything else
2. Move over the windows binary building to the windows executor

2 would be more 'consistent' but 1 would be a simpler change, leaning towards 1 but just thought I'd throw it out there

PR #491, opened 2023-11-06
Add lint rules to detect testing framework assertion usages

### LINKED ISSUE #490: Android Lint check for junit assertions
Consider a new artifact with a lint rule that detects uses of other assertion mechanisms. It probably makes the most sense to ship the rule in a separate artifact in the assertk maven group. Quick Fixes would be appreciated also—especially if they could be batch applied.

- `kotlin.test` assertions
- `org.junit.Assert`
- `org.junit.jupiter.api.Assertions`

The central value in my mind is in discouraging use of assertion mechanisms that ship _with testing frameworks_. But it might also make sense to extend this rule to flag others assertion libraries in common use like Google Truth, Hamcrest, etc.

The benefits are twofold:
1. It will help with incremental adoption of assertk in large projects that currently use other systems.
2. Since the testing frameworks ship with their own assertion libraries, it is very easy for programmers unfamiliar with assertk to reach for those tools instead. A lint check could guide them toward assertk.

jzbrooks:
Adds a few lint checks for detecting assertions from different assertion libraries and from testing frameworks. These lint checks can be used for generic java projects also, they'll just need to apply the android lint gradle plugin.

Unit tests demonstrate the checks, but you can also ad-hoc test the library jar with an environment variable.
https://googlesamples.github.io/android-custom-lint-rules/api-guide.html#publishingalintcheck

Here are some detections that show up from a fresh Android Studio project.
Screenshot 2023-11-05 at 7 59 17 PM

Quick fixes wouldn't be difficult to add, but that can also be done in a separate PR.

The detectors for assertj and google truth are off by default since they aren't included by default with the most popular testing frameworks. I don't think it would be too expensive to run them anyway.

Resolves #490


jzbrooks:
I can run these checks locally. I imagine the machines running the checks in circleci are configured to use an older JDK. I've scanned the log output and have tried to inspect the config but am coming up short. These changes require a newer JDK (17 would do, but using the latest LTS makes sense if possible), but it isn't necessary to change the jvm target for the java artifact. This probably requires upgrading Gradle also.

A couple of other notes:
1. It would be possible to ship these lint checks in an android archive (aar) so that people would get them more or less automatically when they depend on assertk.
2. Publishing configuration needs to be sorted, but I'm happy to do that if y'all are interested.

evant:
Hey! While I think this is useful, I worry about the maintenance burden, particularly with using lint which is known to have breaking changes. How would you feel about moving this to a separate repo that we don't maintain? Would be happy to link to it.

jzbrooks:
That seems fine, although if having android projects to automatically get the checks simply by including assertk as a dependency sounds interesting/valuable it might make more sense here.

I've maintained lint checks for years and haven't encountered a single breaking change (which is not to say it will never happen, just that it hasn't happened to me in quite some time).

That said, building and maintaining lint checks can be onerous—so I get your concern.

evant:
> having android projects to automatically get the checks simply by including assertk as a dependency sounds interesting/valuable it might make more sense here

While I think this is a useful tool, I don't think it should be enabled by default actually, there's many ways one might 
have their project set up where they are purposely using multiple types of assertions.

>  I've maintained lint checks for years and haven't encountered a single breaking change

That's good to hear! I tried making a custom lint ages ago and it was a huge headache, things must of stabilized since then.

> That said, building and maintaining lint checks can be onerous—so I get your concern.

If you check out the recent commit history on this repo it's been slow going. Wish we could spend more bandwidth on it, but that's not in the cards unfortunately, so yeah I still think this is will be better supported externally at least for the time being.

evant:
note: seeing a bunch of
```
'expect'/'actual' classes (including interfaces, objects, annotations, enums, and 'actual' typealiases) are in Beta. You can use -Xexpect-actual-classes flag to suppress this warning. Also see: https://youtrack.jetbrains.com/issue/KT-61573
```
when building locally, may want to add that compiler flag

jzbrooks:
I moved them to https://github.com/jzbrooks/assertk-lint. I plan to publish them at the maven coordinates `com.jzbrooks:assertk-lint:<version>`. We may move them into the Chick-fil-a org eventually.

evant:
Awesome! Put up a pr to link to this in the README, feel free to take a look https://github.com/willowtreeapps/assertk/pull/492

PR #493, opened 2023-11-28
Update Kotlin to 1.9.21

eygraber:
Needed for https://github.com/evant/kotlin-inject/pull/277


buildSrc/src/main/kotlin/assertk.multiplatform.gradle.kts
```diff
@@ -27,19 +31,10 @@ if (libs.versions.assertk.get().endsWith("SNAPSHOT")) {
...
-    "tvosArm64", "tvosX64", "tvosSimulatorArm64",
-    "watchosArm32", "watchosArm64", "watchosX86", "watchosX64", "watchosSimulatorArm64", "watchosDeviceArm64",
-    "androidNativeArm32", "androidNativeArm64", "androidNativeX86", "androidNativeX64",
-)
-
 kotlin {
     jvm()
     js {
+        browser()
         nodejs()
```
JakeWharton, line 38: Why test this and the WASM target twice?

eygraber: Aren't all of them needed for consumers? Especially if https://youtrack.jetbrains.com/issue/KT-47038 is the direction that KMP is headed with this.

JakeWharton: This does not affect consumers in any way today. If the targets split they will split as WASM has, not using these functions. Today it primarily dictates where tests run, and for this library there's no point in running the tests in a browser.

eygraber: It doesn't affect the generated json? Or is that handled at the consumer side?

JakeWharton: Yeah JS is only generated at the very end, by the consumer.

buildSrc/src/main/kotlin/assertk.multiplatform.gradle.kts
```diff
@@ -130,3 +124,21 @@ val detektTest by tasks.registering(Detekt::class) {
 val detekt by tasks.getting {
     dependsOn(detektMain, detektTest)
 }
+
+// Node doesn't publish alpha versions for windows
+if(!DefaultNativePlatform.getCurrentOperatingSystem().isWindows) {
```
evant, line 129: `DefaultNativePlatform` is in an internal package, I think I used `Os.isFamily(Os.FAMILY_WINDOWS)` before

eygraber: That makes more sense 😅

I thought it was something like that. Shouldn't have trusted ChatGPT

eygraber: `Os.isFamily` comes from ant, which I guess isn't available here.

I found `OperatingSystem.current().isWindows` but that is also in an internal package. Seems like the only non internal way of doing this is by using `System.getProperty("os.name")` which is what they use under the hood anyways.

evant: > Os.isFamily comes from ant, which I guess isn't available here

It is, but I never had issues importing it in the past? Anyway using `System.getPropery("os.name")` is fine too.

eygraber: Guess my IDE was acting funny. Using the FQN worked, and now it's working with an import. Pushing change now.

evant (approved):

Issue #495, opened 2023-11-29
Some useablity improvements inspired by truth


evant:
Truth has a [puzzlers](https://truth.dev/comparison) section that includes some cases where assertions can be confusing. There's a few of these where assertk falls into the same trap and can be improved.

problem: `isSameAs`/`isNotSameAs` naming is not quite clear enough that it uses reference equality
example:
```kotlin
assertThat(uniqueIdGenerator.next()).isNotSameAs(uniqueIdGenerator.next())
```
solution:
follow truth and rename to `isSameInstanceAs`/`isNotSameInstanceAs`

problem: `containsAll` is ambiguous on what it's actually comparing
example:
```kotlin
assertThat(primaryColors).containsAll(RED, YELLOW, BLUE)
```
solution:
follow truth and rename to `containsAtLeast`

problem: numbers of different types never compare equal
example:
```kotlin
assertThat(s.toShort()).isEqualTo(parsedValues.get(s))
// -> expected:<1[]> but was:<1[]>
```
solution:
I'm a little hesitant to follow truth here and treat byte/short/int/long the same for equality comparisons because it strays from the kotlin equality behavior where `1 == 1.toShort()` would fail to compile. It's also somewhat improved by kotlin's smart literal types because
```kotlin
assertThat(1.toShort()).isEqualTo(1)
```
will treat the second 1 as a short. We can improve the error message through to mark that they are different types.

JakeWharton:
#449 same spirit

PR #496, opened 2023-11-29
Rename isSameAs/isNotSameAs to isSameInstanceAs/isNotSameInstanceAs

evant:
This makes it more clear these functions are about instance identity see: https://truth.dev/comparison

#495

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/any.kt
@@ -28,7 +28,7 @@ fun Assert<Any>.hashCodeFun() = prop("hashCode", Any::hashCode)
 /**
  * Asserts the value is equal to the expected one, using `==`.
  * @see [isNotEqualTo]
- * @see [isSameAs]
+ * @see [isSameInstanceAs]
  */
 fun <T> Assert<T>.isEqualTo(expected: T) = given { actual ->
     if (actual == expected) return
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
Do we care about keeping tests for the old APIs while they still exist?

evant:
The implementation is
```kotlin
fun <T> Assert<T>.isSameAs(expected: T) = isSameInstanceAs(expected)
```
so I don't think it's that important

JakeWharton:
> The implementation is
> 
> ```kotlin
> fun <T> Assert<T>.isSameAs(expected: T) = isSameInstanceAs(expected)
> ```
> 
> so I don't think it's that important

I thought in the past you wanted me to keep tests for the old APIs, but it's absolutely possible I'm confusing this with other projects. I, too, think it's reasonable to omit.

evant:
> > The implementation is
> > ```kotlin
> > fun <T> Assert<T>.isSameAs(expected: T) = isSameInstanceAs(expected)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so I don't think it's that important
> 
> I thought in the past you wanted me to keep tests for the old APIs, but it's absolutely possible I'm confusing this with other projects. I, too, think it's reasonable to omit.

Depends on what it was testing, anything more complicated than forwarding to a function which already has a test I would say to keep, yeah

PR #498, opened 2023-11-29
show bytes as hex on all supported platforms

evant:
- bytes are shown as hex on all platforms that support them (0x00)
- ~shorts are shown as hex (0x0000)~

~note: I'm a bit less sure about the short hex representation, it's a convenient way to make sure it displays uniquely but I don't know if it's the best display format for how they are used in practice~

#495

### DIFF
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/support/support.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/support/support.kt
@@ -4,7 +4,6 @@ package assertk.assertions.support
 internal actual fun displayPlatformSpecific(value: Any?): String {
     return when (value) {
         is Byte -> "0x%02X".format(value)
-        is Short -> "0x%04X".format(value)
         is Float -> "${value}f"
         is Regex -> "/$value/"
         is Class<*> -> value.name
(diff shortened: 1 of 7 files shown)


evant:
If we don't want hex display of shorts #499 may be a better solution there

assertk/src/jvmMain/kotlin/assertk/assertions/support/support.kt
```diff
@@ -4,6 +4,7 @@ package assertk.assertions.support
 internal actual fun displayPlatformSpecific(value: Any?): String {
     return when (value) {
         is Byte -> "0x%02X".format(value)
+        is Short -> "0x%04X".format(value)
```
JakeWharton, line 7: Shorts in hex is a little weird I think. Most short usage in my experience is for a bitwise protocol, but it still is usually a decimal value.

JakeWharton (approved):

assertk/src/jvmMain/kotlin/assertk/assertions/support/support.kt
evant: yeah the more I'm thinking about it the more I'm against it, think I'm going to adjust this one to only the show bytes as hex on all platforms part

PR #499, opened 2023-11-29
print disambiguation message if display is the same

### LINKED ISSUE #449: Better error messages when things are not equal, but the result of `toString()` is the same.
Consider:

```kotlin
import androidx.compose.ui.text.AnnotatedString
...
        assertThat(AnnotatedString("asdf")).isEqualTo("asdf")
```

This correctly fails, but the error message can be a bit confusing. 

as of `0.25`, the error message is: 
```
AssertionFailedError: expected:<["asdf"]> but was:<[asdf]>
```

A couple of thoughts. 

1. add the quotes for any `CharSequence` not just `String`. so that  `AnnotatedString` would print as `"asdf"` (with quotes)
2. Add type information to the failure message when the `toString()` result is the same, but they are not equal? 
   ```
   AssertionFailedError: expected:<["asdf"]> (type `java.lang.String`) but was:<[asdf]> (type `androidx.compose.ui.text.AnnotatedString`)
   ```

evant:
if the types are different, add them to the error message, otherwise just print out that they don't compare equal

Fixes #449

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt
@@ -31,6 +31,7 @@ internal fun display(value: Any?): String {
             prefix = "{",
             postfix = "}",
             transform = { (k, v) -> "${display(k)}=${display(v)}" })
+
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
Good start. I can spin the individual other cases from that issue into their own.

PR #500, opened 2023-11-29
Add assume function to abort tests instead of failing them

### LINKED ISSUE #432: Missing assumptions
Hi! I'm comparing assertk to assertj and bumped into innability to find assumptions (assumeThat).

Do they exist or are they not yet implemented?

IMO those are quite useful, and should be provided ootb.

evant:
jvm only, uses open4j's TestAbortedException under the hood so works on junit5 or other compatible testing framework (does not work on junit4)

Fixes #432


assertk/src/jvmMain/kotlin/assertk/assume.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Aborts the test instead of failing it, this gives you a way to skip tests based on a dynamic assertion.
+ *
+ * ```
+ * // only run test on windows
+ * assume {
+ *     assertThat(System.getProperty("os.name")).startsWith("Windows")
+ * }
+ * ```
+ */
+fun assume(f: () -> Unit) {
```
evant, line 13: Should we make this inline like other similar functions to allow suspend functions to be called inside?

assertk/src/jvmMain/kotlin/assertk/failure.kt
```diff
@@ -18,3 +21,21 @@ internal actual inline fun Throwable.addSuppressed(error: Throwable) {
 }
 
 internal actual inline fun Throwable.isOutOfMemory(): Boolean = this is OutOfMemoryError
+
+internal object AssumptionFailure : Failure {
+    override fun fail(error: Throwable) {
+        failWithNotInStacktrace(
+            TestAbortedException(
```
evant, line 28: This will treat every exception as an abort, should we only catch `AssertionFailedError` exceptions so that other programming errors continue to fail the test?

evant: Added additional tests, It seems like https://github.com/willowtreeapps/assertk/pull/503 handles this

assertk/src/jvmMain/kotlin/assertk/assume.kt
evant: done

evant:
update: I'm worried about fact that this won't work in junit4 projects which is still very common on android unfortunately, will have to think about dealing with this, would documentation be enough?

PR #502, opened 2023-11-29
add map assertion doesNotContainKey

nukeforum:
Just adding a convenience assertion `Map::doesNotContainKey`. While existing assertions on `Map` can achieve the same functionality, having to pass an arbitrary value in addition to the key to be checked is overhead this PR endeavors to eliminate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added `doesNotContainKey` assertion for `Map`
 
 ## [0.27.0] 2023-09-13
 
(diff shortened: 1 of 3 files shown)


jzbrooks:
In the same way that `isNull` is a terminal assertion while `isNotNull` is both an assertion and transformation, `doesNotContainKey` is a terminal assertion while `key` is an assertion and transformation.

JakeWharton:
I would not expect this to be a terminal operation. It operates on the contents of the map which does not preclude additional assertions on it.

nukeforum:
> ... which does not preclude additional assertions on it.

@JakeWharton 
Isn't that what the `.all {}` extension is for? I'm actually pretty new to assertK, so I could use some context to understand your feedback. 🙏

JakeWharton:
Sorry yeah you're right. I've been working on a different testing library in Rust and it's been screwing up my knowledge of AssertK now multiple times today 🤦‍♂️

evant:
yeah, https://github.com/willowtreeapps/assertk/discussions/346

evant (approved):

evant:
This does make me wonder if it's worth renaming `key()` to `containsKey()`, would make it more clear it's an assertion itself, but maybe it makes it slightly less clear that chaining targets the specific value, kinda 50/50 on it.

Issue #511, opened 2024-01-06
assertFailsWith<SomeExceptionType> { ... }


swankjesse:
[This PR](https://github.com/square/okhttp/pull/8177) migrated about 500 exception tests in OkHttp from this form:

```kotlin
try {
  doThing()
  fail()
} catch (expected: SomeSpecificExceptionType) {
}
```

to this form:

```kotlin
assertFailsWith<SomeSpecificExceptionType> {
  doThing()
}
```

I had to use kotlin.test’s assertFailsWith() even though I generally prefer assertk’s APIs, because assertk doesn’t have a function that lets me expect a failure of a specific type in one go.

I know I could have done this:

```kotlin
assertFailure {
  doThing()
}.isInstanceOf<SomeSpecificExceptionType>()
```

But it’s more code, and I’m in the code deleting business.

PR #525, opened 2024-03-27
Adding new assertFailureWith

robson-alecio:
Adding the new `assertFailureWith` to assert exception with a specific type and verify properties of this type.

### DIFF
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -190,3 +190,17 @@ inline fun assertFailure(f: () -> Unit): Assert<Throwable> {
     }
     fail("expected failure but lambda completed successfully")
 }
+
+/**
+ * Asserts that the given block will throw an exception with expected type rather than complete successfully.
+ */
+inline fun <reified T: Throwable> assertFailureWith(f: () -> Unit): Assert<T> {
+    @Suppress("TooGenericExceptionCaught") // Intentionally capturing all exceptions.
+    try {
+        f()
+    } catch (t: Throwable) {
+        if (t !is T) fail("2expected failure to be type of ${T::class}")
+        return assertThat(t)
+    }
+    fail("expected failure but lambda completed successfully")
+}
--- a/assertk/src/commonMain/kotlin/assertk/assert.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assert.kt
@@ -199,7 +199,7 @@ inline fun <reified T: Throwable> assertFailureWith(f: () -> Unit): Assert<T> {
     try {
         f()
     } catch (t: Throwable) {
-        if (t !is T) fail("2expected failure to be type of ${T::class}")
+        if (t !is T) fail("expected failure to be type of ${T::class}")
         return assertThat(t)
     }
     fail("expected failure but lambda completed successfully")
--- a/assertk/src/commonMain/kotlin/assertk/assertions/throwable.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/throwable.kt
@@ -1,8 +1,8 @@
 package assertk.assertions
 
 import assertk.Assert
-import assertk.ValueAssert
 import assertk.all
+import kotlin.reflect.KProperty1
 
 /**
  * Returns an assert on the throwable's message.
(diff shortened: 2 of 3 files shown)


assertk/src/commonMain/kotlin/assertk/assert.kt
```diff
@@ -190,3 +190,17 @@ inline fun assertFailure(f: () -> Unit): Assert<Throwable> {
...
+
+/**
+ * Asserts that the given block will throw an exception with expected type rather than complete successfully.
+ */
+inline fun <reified T: Throwable> assertFailureWith(f: () -> Unit): Assert<T> {
+    @Suppress("TooGenericExceptionCaught", "InstanceOfCheckForException") // Intentionally capturing all exceptions.
+    try {
+        f()
+    } catch (t: Throwable) {
+        if (t !is T) fail("expected failure to be type of ${T::class}")
```
JakeWharton, line 202: I would either:
- Rethrow exceptions of the wrong type allowing the test to fail normally
- Use `assertFailure(f).isInstanceOf(T::class)` as the whole implementation here which will give you a better message and include the exception as the cause

assertk/src/commonTest/kotlin/test/assertk/AssertFailureTest.kt
```diff
@@ -24,6 +21,30 @@ class AssertFailureTest {
...
     }
 
+    @Test
+    fun failure_with_expected_type_and_additional_properties() {
+        assertFailureWith<DummyException> {
+            throw DummyException(12, "night")
+        }.hasProperties(
+            DummyException::value to 12,
+            DummyException::tag to "night"
+        )
```
JakeWharton, line 31: In general the library does not use its own assertion mechanism to test itself since that could lead to bugs not getting caught by the tests.

For this case, however, I don't think it's worthwhile asserting at such a granular level. The only thing we care about here is that the caught exception is the same as the one thrown, so that's all we need to test (the above test also does this pattern):
```suggestion
        val expected = throw DummyException(12, "night")
        val actual = assertFailureWith<DummyException> { throw expected }
        assertSame(expected, actual.valueOrFail)
```

Goooler:
This addresses #511.

assertk/src/commonMain/kotlin/assertk/assert.kt
robson-alecio: I choose the second approach.

assertk/src/commonTest/kotlin/test/assertk/AssertFailureTest.kt
robson-alecio: I did this and put the tests for `hasProperties` in a better place... I think 😆

assertk/src/commonMain/kotlin/assertk/assert.kt
```diff
@@ -190,3 +191,13 @@ inline fun assertFailure(f: () -> Unit): Assert<Throwable> {
...
 }
+
+/**
+ * Asserts that the given block will throw an exception with expected type rather than complete successfully.
+ */
+inline fun <reified T: Throwable> assertFailureWith(f: () -> Unit): Assert<T> {
+    val assertFailure = assertFailure(f)
+    assertFailure.isInstanceOf(T::class)
+    @Suppress("UNCHECKED_CAST")
+    return assertFailure as Assert<T>
```
JakeWharton, line 202: This returns a specialized `Assert` with the new type
```suggestion
    return assertFailure.isInstanceOf(T::class)
```

assertk/src/commonMain/kotlin/assertk/assertions/throwable.kt
```diff
@@ -62,5 +62,16 @@ fun Assert<Throwable>.hasRootCause(cause: Throwable) {
     }
 }
 
+/**
+ * Asserts the throwable with a specific type have the expected properties for it.
+ */
+fun <T: Throwable> Assert<T>.hasProperties(vararg pairs: Pair<KProperty1<T, Any>, Any>) {
```
JakeWharton, line 68: This is unrelated to `assertFailsWith`. It's not clear why it's type parameter is bound to `Throwable` subtypes since it's a general-purpose utility. I would separate it into its own PR, or even start with an issue discussing the use case and why regular `.all { }` in insufficient.

JakeWharton (approved):
I would prefer this variant of `assertFailure`, but it was asked to be omitted when that function was first introduced. You'll have to wait to hear from a maintainer on #511

PR #542, opened 2024-07-09
Add Path `doesNotExist` assertion

jisungbin:
`Assert<Path>.exists()` exists, but its negation, `Assert<Path>.doesNotExist()`, does not. Add `doesNotExist()` to Path for clean negation assertions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - renamed `prop` to `having` as part of effort to unify API naming, old name is deprecated.
 - renamed `suspendCall` to `having` as part of effort to unify API naming, old name is deprecated.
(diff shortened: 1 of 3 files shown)


evant (approved):
Looks good! fyi macos build is failing because the resource class needs to be updated (see https://github.com/evant/kotlin-inject/pull/410/commits/6d896910d8e66ac56457e0ffa91bb0ed61a16d79). I'll get to it eventually but feel free to do it here in a separate commit if you want.

jisungbin:
> Looks good! fyi macos build is failing because the resource class needs to be updated (see [evant/kotlin-inject@6d89691](https://github.com/evant/kotlin-inject/commit/6d896910d8e66ac56457e0ffa91bb0ed61a16d79)). I'll get to it eventually but feel free to do it here in a separate commit if you want.

I fixed the CI with the commit you referenced, and now the macos build passes! 🥳

PR #546, opened 2024-09-29
Add collection filtering assertions analagous to instanceOf

jzbrooks:
This would be pretty handy when dealing with collections of sealed typed objects. I'd use it a lot!


CHANGELOG.md
```diff
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - renamed `prop` to `having` as part of effort to unify API naming, old name is deprecated.
 - renamed `suspendCall` to `having` as part of effort to unify API naming, old name is deprecated.
-- added `doesNotExist` assertions to `Path`.
+
+### Added
+- Added `doesNotExist` assertions to `Path`
```
jzbrooks, line 14: I moved this note into the added section for consistency

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
```diff
@@ -131,6 +133,34 @@ internal fun MutableList<*>.removeFirst(value: Any?) {
...
+ *
+ * ```
+ * assertThat(listOf<Any>("one", "two", 1)).containsInstanceOf<String>().each {
+ *   it.hasLength(3)
+ * }
+ * ```
+ */
+inline fun <reified T> Assert<Iterable<*>>.containsInstanceOf(): Assert<List<T>> {
+    return transform("contains subtype of ${T::class}") { actual ->
+        actual.filterIsInstance<T>().also {
```
JakeWharton, line 147: I would use `any` here so that it returns `true` as soon as a single instance is found rather that needing to perform a full iteration and allocate a new list.

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
```diff
@@ -131,6 +133,34 @@ internal fun MutableList<*>.removeFirst(value: Any?) {
...
+
+/**
+ * Asserts the collection does not contain an instance of a given type.
+ *
+ * ```
+ * assertThat(listOf<Any>("one", "two", 1)).doesNotContainInstanceOf<Double>()
+ * ```
+ */
+inline fun <reified T> Assert<Iterable<*>>.doesNotContainInstanceOf() = given { actual ->
+    if (actual.filterIsInstance<T>().isNotEmpty()) expected("to contain no instances of ${T::class} but was $actual")
```
JakeWharton, line 161: Similarly, `none` here.

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
jzbrooks: The sub-list allocation is necessary to provide a way to continue making assertions against the sublist. No?

JakeWharton: Ah, yeah, I missed the new list was propagated as a return value. I'm a bit hesitant about the API shape using a verb "contains" while the return value is a full filtering.

jzbrooks: Yeah, agreed. I noticed the `having` discussion after I pushed this. Would `havingInstancesOf` be better? Maybe the negation would be `doesNotHaveInstancesOf`?

JakeWharton: I guess I'm a bit torn on whether this makes sense as a dedicated operator since it combines a transform with an assertion in a way that I don't think has precedent from other operators. We'll see what the actual maintainers think, though.

jzbrooks: I'd argue there's precedent with `isNotNull`. It's a transform and an assertion.

jzbrooks: Fair. I'd find a shorthand for the transformation basically just as helpful.

jzbrooks:
#554 might be relevant

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
jzbrooks: A transformation without assertion might set a (good) precedent for other collection operators.
@evant, could you weigh in when you have some time?

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
```diff
@@ -131,6 +133,34 @@ internal fun MutableList<*>.removeFirst(value: Any?) {
...
+ * Asserts the collection contains at least one instance of a given type.
+ *
+ * ```
+ * assertThat(listOf<Any>("one", "two", 1)).containsInstanceOf<String>().each {
+ *   it.hasLength(3)
+ * }
+ * ```
+ */
+inline fun <reified T> Assert<Iterable<*>>.containsInstanceOf(): Assert<List<T>> {
+    return transform("contains subtype of ${T::class}") { actual ->
```
jzbrooks, line 146: This message should probably just be `contains instance of ${T::class}` per Liskov.

jzbrooks:
Along the lines of the `key` argument in #557, is there a precedent for or idea of what the negation of `having` should be called? `notHaving`?

jzbrooks:
Alternatively `<T> havingInstancesOf` could make sense as just a transformation.

assertk/src/commonMain/kotlin/assertk/assertions/iterable.kt
jzbrooks: `single` is also an assertion and transformation of the same kind

PR #552, opened 2024-10-25
Fix `isDataClassEqualTo` not working with `data object`s

### LINKED ISSUE #551: `isDataClassEqualTo` passes when given different `data object`s
This test should fail, but it passes:
```kotlin
data object Example
data object AnotherExample

@Test
fun test() {
    assertThat(Example).isDataClassEqualTo(AnotherExample)
}
```
When using a normal `object`, it does fail since they aren't data classes:
```kotlin
object Example
object AnotherExample

@Test
fun test() {
    assertThat(Example).isDataClassEqualTo(AnotherExample)
    /*
    only supports data classes
        java.lang.IllegalArgumentException: only supports data classes
    */
}
```
And the different `data object`s correctly fail when using `isEqualTo`:
```kotlin
data object Example
data object AnotherExample

@Test
fun test() {
    assertThat(Example).isEqualTo(AnotherExample)
    /*
    Expected :[Another]Example
    Actual   :[]Example

    org.opentest4j.AssertionFailedError: expected:<[Another]Example> but was:<[]Example>
    */
}
```
I'd expect this to fail similarly to using `isEqualTo` or with a message that they're not supported. The current behavior gives the impression of a passing test.

VirtualParticle:
This fixes #551 by going right to `isEqualTo` if the passed value is a `data object`.

### DIFF
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
@@ -83,7 +83,7 @@ fun <T : Any> Assert<T>.isDataClassEqualTo(expected: T) = given { actual ->
 private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?): Unit = given { actual ->
     if (actual == expected) return
     val compareProps = actual != null && expected != null
-    if (compareProps && kclass != null && kclass.isData) {
+    if (compareProps && kclass != null && kclass.isData && kclass.objectInstance == null) {
         for (memberProp in kclass.memberProperties) {
             @Suppress("UNCHECKED_CAST")
             val force = memberProp as KProperty1<T, Any?>
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
@@ -82,8 +82,8 @@ fun <T : Any> Assert<T>.isDataClassEqualTo(expected: T) = given { actual ->
 
 private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?): Unit = given { actual ->
     if (actual == expected) return
-    val compareProps = actual != null && expected != null
-    if (compareProps && kclass != null && kclass.isData && kclass.objectInstance == null) {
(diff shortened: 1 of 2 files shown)


assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -82,7 +82,7 @@ fun <T : Any> Assert<T>.isDataClassEqualTo(expected: T) = given { actual ->
 private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?): Unit = given { actual ->
     if (actual == expected) return
     val compareProps = actual != null && expected != null
-    if (compareProps && kclass != null && kclass.isData) {
+    if (compareProps && kclass != null && kclass.isData && kclass.objectInstance == null) {
```
VirtualParticle, line 85: This seemed like the best way to check if `kclass` is a `data object`, but there could be a better way.

evant: could we check the types directly? This should always fail if `expected::class != actual::class` right?

VirtualParticle: Sure, how does this look?
```kotlin
    val compareProps = actual != null && expected?.let { it::class } == kclass
    if (compareProps && kclass != null && kclass.isData) {
```

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -81,7 +81,7 @@ fun <T : Any> Assert<T>.isDataClassEqualTo(expected: T) = given { actual ->
 
 private fun <T> Assert<T>.isDataClassEqualToImpl(expected: T, kclass: KClass<*>?): Unit = given { actual ->
     if (actual == expected) return
-    val compareProps = actual != null && expected != null
+    val compareProps = actual != null && expected?.let { it::class } == kclass
     if (compareProps && kclass != null && kclass.isData) {
```
VirtualParticle, line 85: This technically removes the `expected != null` check, but it's essentially still done since we check for equality with `kclass` which is then later checked not to be `null`.

evant (approved):

PR #553, opened 2024-10-28
Widen boolean asserts to operate on nullable types

### LINKED ISSUE #550: Relax bounds of built-in value-asserting functions.
I'm migrating a couple thousand test cases from Truth, and another thing that is coming up similar to https://github.com/willowtreeapps/assertk/issues/393#issuecomment-2428206350 is that calls to `isTrue()` need to become `isNotNull().isTrue()` on nullable subjects. In general this makes sense! However, the fact that I can call `.isEqualTo(true)` _without_ the `isNotNull()` check just feels bad. It's shorter! Why wouldn't I _always_ then do that?

I fully understand and am justifying calls to `isNotNull()` before things like `startsWith()` which operate on the actual value and therefore need to null-check it first. But `isTrue()` and `isFalse()` seem like they should be equivalent to `isEqualTo(true)` and `isEqualTo(false)`, just maybe with a better message. That makes me think they should operate on `Boolean?` and not just `Boolean`.

JakeWharton:
This behavior brings it more line with how '==' behaves in the language.

Closes #550.


evant (approved):

JakeWharton:
Thanks for fixing the infra. Didn't know what was going on there.

evant:
Me neither, I just re-ran the job 😂

Issue #393, opened 2021-11-11
`isInstanceOf` should be permitted to call on a nullable


hakanai:
I tried to write the following today:

```kotlin
val extension: Any? = ...;

assertThat(extension).isInstanceOf(WarningsOptionsExtension::class.java)
```

I was surprised that this was not permitted, because the equivalent in Hamcrest would just fail the assertion if the value was `null`.

Workaround is to do it in two steps:

```kotlin
assertThat(extension).isNotNull()
assertThat(extension!!).isInstanceOf(WarningsOptionsExtension::class.java)
```

evant:
You can chain them
```kotlin
assertThat(extension).isNotNull().isInstanceOf(WarningsOptionsExtension::class.java)
```
Hamcrest was written for java which doesn't have nullability in the type system. The explicit `isNull()` assertion makes it so every other assertion doesn't have to check null.

hakanai:
Nice.

JakeWharton:
I would like to propose this be reconsidered.

I'm migrating a few thousand files to AssertK and the need to perform an explicit non-null check before an instance-of check has come up a lot. It feels somewhat asymmetric, especially in comparison to the ability to call `isEqualTo` on a type which is nullable.

Sometimes they appear right next to each other, too:
```kotlin
assertThat(savedState["compositeKeyHash"])
  .isNotNull()
  .isInstanceOf<List<Any?>>()
  .all {
    index(0).isNotNull().isInstanceOf<MutableState<Int>>().prop(MutableState<_>::value).isEqualTo(1)
    index(1).isEqualTo("two")
  }
```

This _could_ be

```kotlin
assertThat(savedState["compositeKeyHash"])
  .isInstanceOf<List<Any?>>()
  .all {
    index(0).isInstanceOf<MutableState<Int>>().prop(MutableState<_>::value).isEqualTo(1)
    index(1).isEqualTo("two")
  }
```

which doesn't feel any less clear. The `is` keyword handles `null` automatically the same way `==` would, so shouldn't `isInstanceOf` handle it the same way `isEqualTo` does?

PR #554, opened 2024-10-28
Widen `isInstanceOf` assert to operate on nullable types

JakeWharton:
This behavior brings it more line with how 'is' behaves in the language.

Refs #393. Refs #550.

### DIFF
--- a/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
@@ -46,12 +46,13 @@ fun <T : Any> Assert<T>.doesNotHaveClass(jclass: Class<out T>) = given { actual 
  * @see [isNotInstanceOf]
  * @see [hasClass]
  */
-fun <T : Any, S : T> Assert<T>.isInstanceOf(jclass: Class<S>): Assert<S> = transform { actual ->
+fun <T : Any> Assert<Any?>.isInstanceOf(jclass: Class<T>): Assert<T> = transform { actual ->
     if (jclass.isInstance(actual)) {
         @Suppress("UNCHECKED_CAST")
-        actual as S
+        actual as T
     } else {
-        expected("to be instance of:${show(jclass)} but had class:${show(actual.javaClass)}")
+        val but = if (actual == null) "was null" else "had class:${show(actual.javaClass)}"
+        expected("to be instance of:${show(jclass)} but $but")
     }
 }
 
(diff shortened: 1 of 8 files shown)


assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -46,12 +46,13 @@ fun <T : Any> Assert<T>.doesNotHaveClass(jclass: Class<out T>) = given { actual
  * @see [isNotInstanceOf]
  * @see [hasClass]
  */
-fun <T : Any, S : T> Assert<T>.isInstanceOf(jclass: Class<S>): Assert<S> = transform { actual ->
+fun <T : Any> Assert<Any?>.isInstanceOf(jclass: Class<T>): Assert<T> = transform { actual ->
```
JakeWharton, line 49: Having two type params here was mostly useless. Kotlin will happily widen `T` to `Any` if the two have no common supertypes. The tests actually rely on this behavior.

evant: Think this was left over from an older version, looks like the common one was already change and this was missed

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
```diff
@@ -46,12 +46,13 @@ fun <T : Any> Assert<T>.doesNotHaveClass(jclass: Class<out T>) = given { actual
...
     if (jclass.isInstance(actual)) {
         @Suppress("UNCHECKED_CAST")
-        actual as S
+        actual as T
     } else {
-        expected("to be instance of:${show(jclass)} but had class:${show(actual.javaClass)}")
+        val but = if (actual == null) "was null" else "had class:${show(actual.javaClass)}"
```
evant, line 54: wondering if we should've done something explicit like this to the `isTrue/isFalse` case to avoid the extra value diff noise. Sorry, missed catching that in the other pr.

evant (commented):
Do we also want `isNotInstanceOf` to take null for the same reason?

assertk/src/jvmMain/kotlin/assertk/assertions/any.kt
JakeWharton: Done (separate commit)

JakeWharton:
> Do we also want `isNotInstanceOf` to take null for the same reason?

Done! I'm not sure how useful this is in practice, but consistency with the positive assertion and the language makes sense.

evant (approved):

PR #555, opened 2024-11-06
Add `isIn(Iterable<T>)` and `isNotIn(Iterable<T>)`

### LINKED ISSUE #549: Version of `IsIn`/`isNotIn` that takes an `Iterable`?
I have my expected values in a constant that's shared, but I have to do `isIn(*values)` with an `arrayOf` constant instead of `isIn(values)` with a `listOf` constant today.

HaydenMeloche:
This adds some convenient methods when asserting something is (and is not) in something else. Previously due to `isIn` and `isNotIn` only accepting `vararg values: T`, this caused a somewhat confusing error.
Before:
```kotlin
@Test
fun demo() {
    val possibleValues = listOf("test")
    assertThat("test").isIn(possibleValues) // expected:<[["test"]]> to contain:<"test">
}
```
As mentioned in #549 to achieve something similar to above, you would need to spread an `Array<T>`
```kotlin
@Test
fun demo() {
    val possibleValues = arrayOf("test")
    assertThat("test").isIn(*possibleValues) // does not error
}
```

Now with `isIn(Iterable<T>)` and `isNotIn(Iterable<T>)`
```kotlin
@Test
fun demo() {
    val possibleValues = listOf("test")
    assertThat("test").isIn(possibleValues) // does not error
}
```
does not error.

fixes #549

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - renamed `suspendCall` to `having` as part of effort to unify API naming, old name is deprecated.
 - added `doesNotExist` assertions to `Path`.
 - the receiver types for `isTrue()`, `isFalse()`, and `isInstanceOf()` have been widened to operate on nullable values.
+- added assertions `isIn(Iterable<T>)` and `isNotIn(Iterable<T>)`
 
 ## [0.28.1] 2024-04-17
 
--- a/assertk/src/commonMain/kotlin/assertk/assertions/any.kt
+++ b/assertk/src/commonMain/kotlin/assertk/assertions/any.kt
@@ -98,10 +98,28 @@ fun <T> Assert<T>.isIn(vararg values: T) = given { actual ->
 }
 
 /**
+ * Asserts the value is in the expected values, using `in`.
+ * @see [isNotIn]
+ */
+fun <T> Assert<T>.isIn(values: Iterable<T>) = given { actual ->
+    if (actual in values) return
+    expected(":${show(values)} to contain:${show(actual)}")
+}
+
+/**
  * Asserts the value is not in the expected values, using `!in`.
  * @see [isIn]
  */
 fun <T> Assert<T>.isNotIn(vararg values: T) = given { actual ->
+    if (actual !in values) return
+    expected(":${show(values)} to not contain:${show(actual)}")
+}
+
+/**
+ * Asserts the value is not in the expected values, using `!in`.
+ * @see [isIn]
+ */
+fun <T> Assert<T>.isNotIn(values: Iterable<T>) = given { actual ->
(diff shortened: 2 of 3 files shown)


JakeWharton (approved):
Thanks! I was going to do this first thing tomorrow. Saved me the trouble!

evant (commented):
oh nice! I was worried about the overloads not working well here, but seems like it might be fine? I think the only edge case might be asserting on a collection but it might not be too confusing. Could you at least add a test case for that?

HaydenMeloche:
> oh nice! I was worried about the overloads not working well here, but seems like it might be fine? I think the only edge case might be asserting on a collection but it might not be too confusing. Could you at least add a test case for that?

@evant I think this case should be covered in the existing tests since I am using `List<T>` which implements `Collection<T>` unless I am misundestanding

evant:
@HaydenMeloche for the subject

HaydenMeloche:
@evant Oh I see, something like:
```kotlin
val value = listOf("test")
val possibleValues = listOf("test")
assertThat(value).isIn(possibleValues)
```

I can add this!

HaydenMeloche:
@evant you're correct. 
```kotlin
    @Test
    fun test() {
        val value = listOf("test")
        val possibleValues = listOf("test")
        assertThat(value).isIn(possibleValues)
    }
```
throws an error: `expected:<["test"]> to contain:<["test"]>`.  

This issue also exists with the existing `isIn(vararg T)` as well. Is this something we want to address in this PR? I'd be happy to do a follow up one to address all variants of `isIn`/`isNotIn` if that's preferred.

JakeWharton:
So the biggest problem with overloading varargs with something which accepts an `Iterable` is that you can have a type (e.g., `A`) which implements `Iterable` of itself (e.g., `A : Iterable<A>`).

I _very_ rarely see this, but it does occur on a fairly widespread type: [`java.nio.file.Path`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html).

This means that today, writing
```kotlin
assertThat(somePath).isIn(otherPath)
```
will change its behavior after this function lands to assert on the result of _iterating_ `otherPath`, not the single-element array which _contains_ `otherPath`.

Now obviously no one should be doing `isIn` with a single element... but I thought it worth pointing out.

I'm still a 👍 on this.

evant:
> Now obviously no one should be doing isIn with a single element... but I thought it worth pointing out

hm, what if we changed the signature to
```
isIn(first: A, second: A, vararg rest: A)
```
and deprecated the old one?

JakeWharton:
Yeah I like it. I'm a bit wary of doing stuff like that in normal libraries because you end up with a bunch of overloads (such as to avoid the varargs allocation when there is only two), but in this case performance or allocation isn't a problem.

JakeWharton:
I will make that change as a separate PR today.

evant (approved):

PR #556, opened 2024-11-13
Force `isIn` and `isNotIn` callers to specify at least two elements

JakeWharton:
From discussions in #555


evant (approved):

JakeWharton:
I forgot to update the change log. Let me do that quickly.

Issue #522, opened 2024-03-19
Consider unifying first-party, `prop`-based helper function names (like `jClass`, `toStringFun`)


JakeWharton:
The current naming is mostly just the property name as the function name. There's `kClass`, collection `size()`s, exception `message()`

This strategy doesn't work for `toString()` or `hashCode`. They get "Fun" suffixes as `toStringFun()` and `hashCodeFun()`, respectively.

This ties in somewhat with #521, and it actually leans me towards that issue choosing "have" so that these could be "has"-prefixed.

```kotlin
assertThat(foo).hasKClass().isEqualTo(Foo::class)
assertThat(array).hasSize().isEqualTo(3)
assertThat(user).hasToString().isEqualTo("User[name=Alice]")
```

grodin:
Once there's an agreement on how to resolve #521, happy to handle this

evant:
There is a subtle distinction there, the 'has*' methods do an assertion themselves while the non-has methods can't fail on their own. One thing I'd want to avoid is someone doing `assertThat(array).hasSize()` by itself not realizing that it will never fail.

grodin:
Perhaps "having" could help here, as it seems less like an assertion: 

```
assertThat(list).havingSize()
```

feels less like a complete sentence.

JakeWharton:
I also think `having` pairs well with `extracting`, should we opt not to change it to something like `eachHaving`.

grodin:
> I also think `having` pairs well with `extracting`, should we opt not to change it to something like `eachHaving`.

Perhaps that's a useful general principle for this kind of thing; a method named with a verb which isn't an assertion should use the continuous ("ing") form of the verb.

That's almost certainly a premature abstraction, and wrong in the details, but I thought it was worth noting down.

evant:
Hm, looking at the [draft mr](https://github.com/willowtreeapps/assertk/pull/537) I'm not sure I'm convinced that adding a `having` prefix to everything is preferable to the status-quo. There are far more methods that the current pattern works for than it doesn't. I think it would still be reasonable to rename `toStringFun` and `hashCodeFun` to `havingToString` and `havingHashCode` respectively, but thinking it may be better to keep the rest the same. Thoughts?

JakeWharton:
What are your primary examples of occurrences you don't like.

evant:
From a wider context we are introducing api churn. Which in-itself isn't a bad thing but does mean that there should be a good reason for it.

[...]

PR #557, opened 2024-11-15
Add transformers `Map<K,V>.havingKeys()` and `Map<K,V>.havingValues()`

### LINKED ISSUE #527: Assert on map should have keys() and values() functions
There are good assertions on individual items in maps but not a clean way to do assertions on all keys or all values in a map. So i propose the following methods be added:
```
fun <K> Assert<Map<K, *>>.keys() : Assert<Set<K>> = ...

fun <V> Assert<Map<*, V>>.values() : Assert<Collection<V>> = ...
```

HaydenMeloche:
This adds some transformers allowing you to utilize `Collection` asserts on parts of a map.

Example for keys:
```kotlin
assertThat(mapOf("key" to "value")).havingKeys().containsOnly("key")
```
Example for values: 
```kotlin
assertThat(mapOf("key" to "value")).havingValues().containsOnly("value")
```

Given that there is already a `Map<K,V>.isNotEmpty()` I thought it made sense to have these transformers throw an error if their corresponding parts have no values.

fixes #527

### DIFF
--- a/assertk/src/commonTest/kotlin/test/assertk/assertions/MapTest.kt
+++ b/assertk/src/commonTest/kotlin/test/assertk/assertions/MapTest.kt
@@ -287,4 +287,34 @@ class MapTest {
         assertEquals("expected [subject] to have key:<\"wrong\">", error.message)
     }
     //endregion
+
+    //region havingKeys
+    @Test
+    fun havingKeys_empty_list_fails() {
+        val error = assertFailsWith<AssertionError> {
+            assertThat(emptyMap<String, String>()).havingKeys()
+        }
+        assertEquals("expected map to not be empty", error.message)
+    }
+
+    @Test
+    fun havingKeys_assertion_passes() {
+        assertThat(mapOf("key" to "value")).havingKeys().containsOnly("key")
+    }
+    //endregion
+
+    //region havingValues
+    @Test
+    fun havingValues_empty_list_fails() {
+        val error = assertFailsWith<AssertionError> {
+            assertThat(emptyMap<String, String>()).havingKeys()
+        }
+        assertEquals("expected map to not be empty", error.message)
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


JakeWharton:
> I thought it made sense to have these transformers throw an error if their corresponding parts have no values.

Do any other "having"-style functions also perform an implicit assertion?

Really my only criticism of this approach is that there's something pure in the "having" suite of functions being purely transformational in contrast to actual validation functions. There's nothing _strictly_ wrong with doing `assertThat(map).havingKeys().isEmpty()`, after all, even if it's a bit redundant.

JakeWharton:
I did find some examples, such as `key` on a map (which should become `havingKey` if #522 moves forward). So retaining the implicit assertion here seems fine 👍 .

HaydenMeloche:
@JakeWharton I admittedly was on the fence here (and I'm definitely open to removing the assertion). `Map<K,V>.key()` was ultimately the reason I decided to leave in the assertion

jzbrooks:
This seems analogous to #546 (function names in that PR notwithstanding).

assertk/src/commonTest/kotlin/test/assertk/assertions/MapTest.kt
```diff
@@ -287,4 +287,34 @@ class MapTest {
...
+    fun havingKeys_assertion_passes() {
+        assertThat(mapOf("key" to "value")).havingKeys().containsOnly("key")
+    }
+    //endregion
+
+    //region havingValues
+    @Test
+    fun havingValues_empty_list_fails() {
+        val error = assertFailsWith<AssertionError> {
+            assertThat(emptyMap<String, String>()).havingKeys()
```
emartynov, line 310: Should it be `havingValues()`?

emartynov (changes requested):
One comment and the rest is LGTM

PR #559, opened 2024-12-03
Add `asAssert` extensions

Goooler:
This is handy for simplifying things in https://github.com/GradleUp/shadow/blob/6c1920568d4c9c98ed5bb86ebf106a92ff63ef01/src/funcTest/groovy/com/github/jengelman/gradle/plugins/shadow/util/file/TestFile.java#L280-L298

to

```kt
  fun assertExists(): TestFile = apply {
      asAssert().exists()
  }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,14 +8,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Add `asAssert` extensions for simplifying `assertThat` usages.
+- Add `doesNotExist` assertions to `File` and `Path`.
+- Add assertions `isIn(Iterable<T>)` and `isNotIn(Iterable<T>)`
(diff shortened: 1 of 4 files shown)


JakeWharton:
This seems a bit specialized for what is effectively saving four characters on `assertThat(this)`.

Goooler:
They are convenient for sequence invocations like `true.assertThis().isTrue()`, and as extensions of what could be a part of idiomatic Kotlin usages.

JakeWharton:
Hmm I would expect an extension in that form to be named `asAssert()` to fit in with other extension-based conversion functions.

Goooler:
Makes sense!

evant:
Sorry, this is to much of "multiple ways to do the same thing" that I want to avoid. There's other assertion libraries out there if you prefer a different style of assertion.

PR #561, opened 2025-01-12
Add containsKey assertion for Map

Goooler:


jzbrooks:
[The key assertion does this](https://github.com/willowtreeapps/assertk/blob/a69843f4452d2cf83bb6a787b138f8d90cfc3f90/assertk/src/commonMain/kotlin/assertk/assertions/map.kt#L162-L169). What's the value add?

Goooler:
Inverted for 

https://github.com/willowtreeapps/assertk/blob/a69843f4452d2cf83bb6a787b138f8d90cfc3f90/assertk/src/commonMain/kotlin/assertk/assertions/map.kt#L118-L124

jzbrooks:
The `key` assertion asserts that the map contains a key and transforms the assert into an assertion on the value. You don't have to use the latter part of that functionality if you don't want to. The former does the same thing the proposed assertion does.

PR #564, opened 2025-02-24
Check the Gradle distribution checksum

TheMrMilchmann:
It is generally a good idea to ensure the integrity of the Gradle distribution to prevent MITM attacks when downloading. (See https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification)

Keep in mind that this has to be updated alongside the distribution link.


jzbrooks (approved):
Looks good—the best way to upgrade these things is via the `wrapper` task anyway, which automatically updates this.

```
% shasum -a 256 ~/Downloads/gradle-8.4-bin.zip                                                                                                                    (master)
3e1af3ae886920c3ac87f7a91f816c0c7c436f276a6eefdb3da152100fef72ae  /Users/justin/Downloads/gradle-8.4-bin.zip
```

PR #582, opened 2026-08-29
Add BooleanArray array assertions via the array template.

mpawliszyn:
Note that the documentation uses numbers as examples but they are still representative of the expected behaviour.

Based on the forensics I did BooleanArray appears to be simply missed during development. Also note that I had to adapt the tests to the boolean values rather than using the number array templating directly.

I added documentation that references the templated tests. Let me know if you want me to add a back-reference in the templates as well. I could see wanting that going either way.

I did run all the tests locally on my mac as best I could on everything that does not require xcode.


mpawliszyn:
😓 Finally got a green build. 

I'll split out the CI fixes too.

It may make sense to move off of the canary node which I think also requires a move to kotlin 2.0 (although we should be able to also set `languageVersion = "1.9"` )

